PR #199, opened 2019-08-05
V7.5

verenich:


raboczi:
All my code is in, including the fix I passed to Bruce.

PR #314, opened 2020-05-21
AP-1481 Reorganize about

dreamind:


dreamind:
> It's not so great that MenuController in the portal treats menus named "About" and "Account" specially, but I appreciate that we'd otherwise have to add some API for plugins to request their menu items to be on the right side of the menubar.

Yes, I agree. We need a better way to handle this. Probably defining more comprehensive structure in site.properties (or a YAML-version of it) would make a better choice (in the future). I will create a placeholder ticket for this.

PR #392, opened 2020-08-07
AP-1414 Fix next trace issue

dreamind:


raboczi (commented):
In lines 586 and 604, it seems weird that you divide traceS by 1000 and then immediately multiply it by 1000 in the only(?) place it's subsequently used.  Maybe just have traceMs instead?

raboczi (approved):

PR #424, opened 2020-09-03
Feature/ap 776 configurable ldap

raboczi:
This modification allows LDAP login to be configured without having to recompile the portal bundle.  Specifically:

- the UserSessionManager.initializeUser method which copies users from LDAP is now configurable from site.properies rather than being hardcoded for the University of Melbourne "Centaur" LDAP

- login.conf and portalContext-security.xml are no longer embedded within the portal bundle, but are instead located in virgo-tomcat-server-3.6.4.RELEASE/configuration/.

- the ApromoreCore README instructions for configurating LDAP have been updated accordingly


raboczi:
This PR was supposed to be against development, not master.

PR #426, opened 2020-09-04
* Clean up unused artifacts

nolantellis:
* Cleanup unused plugins
* emoved Source code jar plugin
* Removed flyway Repo
* Removed Resource plugin as not used as custom plugin

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,6 +24,7 @@
 
 # Maven artifacts (i.e. object files)
 **/target/*
+**/.settings/*
 
 # Tomcat artifacts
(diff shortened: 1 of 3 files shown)


raboczi:
Builds and runs, and by inspection doesn't remove anything that isn't dispensable. The only thing I'd want to confirm is that the author of the Flyway modifications doesn't require them anymore.  I'll have to crawl the history to determine who that might be.

raboczi (approved):
Confirmed with Ivo that's happy to see the Flyway dependencies removed, and I've confirmed by inspection and by building locally that nothing of value has been lost.

PR #447, opened 2020-09-16
Added liquibase changes and test for convinience

nolantellis:


raboczi (approved):
If the H2 .db files can be configured into the target directory which Maven creates as ${project.build.outputDirectory}, that might remove the need for hiding *.db in the .gitignore.

I'm satisfied this is good enough to commit, regardless.

nolantellis:
> If the H2 .db files can be configured into the target directory which Maven creates as ${project.build.outputDirectory}, that might remove the need for hiding *.db in the .gitignore.
> 
> I'm satisfied this is good enough to commit, regardless.

I added the h2 file to test/resource/directory as well as in git ignore. The file will be recreated, Not sure if target directory will exist during the test phase.

PR #465, opened 2020-09-25
Change git plugin to work only for apromore core and not all module

nolantellis:


raboczi (approved):
As long as the About... plugin continues to work, this is fine.

PR #487, opened 2020-10-01
AP-2311 Replace messagebox with more subtle notification

dreamind:


raboczi (approved):
Built it locally and it works; code seems tidy enough.  Approved!

Mightn't hurt to make the Notify class "abstract final" since nobody should be creating an instance from it or extending it, and then the "final" on the methods is implied.

franklma (approved):

Apromore-Core-Components/Apromore-Portal/src/main/java/org/apromore/portal/common/Notify.java
```diff
@@ -0,0 +1,43 @@
...
+
+    public static final void info(String message) {
+        Clients.evalJavaScript("Ap.common.notify('" + message.replace("'", "\\'") + "','info');");
+    }
+
+    public static final void error(String message) {
+        Clients.evalJavaScript("Ap.common.notify('" + message.replace("'", "\\'") + "','error');");
+    }
+
+}
```
brucenguyen1: "Notify" is not a right name for a class as it's confusing with a method name, e.g. Ap.common.notify() is a method used in the javascript code. Suggest to change it to Notification, MessageBox, or a noun. Naming is a simple coding step but has a lasting effect on the code readability. It seems not a big deal here in this particular Notify.java file but I would suggest we try to stick to good coding common sense.

brucenguyen1: Another comment: is there a good reason to change from using ZK's built-in facilities to our own implementation? I guess it's a yes of course and you can write your justification here.

brucenguyen1: On the other hand, do we have a unit test for this new class? For example, the message could have special characters, HTML tags. What happens if type is an arbitrary value? The unit tests would be a good place to test and also put these rules as a kind of documentation for anybody to understand how to use this utility which seems to be a good one for us.

raboczi:
A trivial but worthwhile optimization: implement Notify.info and Notify.error as calls to Notify.show so that the Javascript string escaping is only done in one place rather than repeated.

Apromore-Core-Components/Apromore-Portal/src/main/java/org/apromore/portal/common/Notify.java
dreamind: A similar style of this contemporary notification is available only in ZK 9. This is just a class for abstracting the ZK implementation later on. Moreover, this supports HTML snippets like bold important text.

raboczi: If ZK 9 has features like this we actually want, it might be worth our while to do the upgrade.  I'm certainly open to investigating how much effort it'd take.

dreamind:
Unit test is added. Most of the suggestions have been included the updates.

nolantellis (approved):

dreamind:
Add enumeration for notification type

PR #491, opened 2020-10-04
AP-2234 Immediately process search on combobox select

dreamind:


nolantellis (approved):

Apromore-Core-Components/Apromore-Portal/src/main/java/org/apromore/portal/dialogController/SimpleSearchController.java
```diff
@@ -107,6 +107,7 @@ public void onEvent(Event event) throws Exception {
         previousSearchesCB.addEventListener("onSelect", new EventListener<Event>() {
             public void onEvent(Event event) throws Exception {
                 setVisibility(clearSearchBtn, true);
+                processSearch();
```
brucenguyen1, line 110: As this system-wide "search" function requires extensive queries affecting performance, it is not recommended to perform immediately after the keyword is selected.

brucenguyen1: Another usability issue to consider is: as soon as a previous keyword is selected (e.g. "sepsis diagnosis_CRP"), the system searches and returns a result; however, this previous keyword is only selected to be refined, the actual keywords the user wants to use is: "sepsis case_short diagnosis_CRP". This action thus results in an unnecessary system search (could involve a waiting hourglass) which is an inefficient UI choice.

brucenguyen1 (changes requested):

dreamind:
Close the PR for now

PR #492, opened 2020-10-04
Merge Feature/AP-2300 to development

odaghagheleh:
Development merged to AP-2300 and Virgo started successfully


raboczi (approved):
The size of this commit is a bit overwhelming to review, but we've done a few passes and it seems to build and run acceptably.

PR #500, opened 2020-10-06
Seperate Cache from database package for future use

nolantellis:
This is Temporary movement.
Moved caching stuff out of logRepo to a Temp Cache service.

This will help in moving Caching as well as db out of manager.


Apromore-Core-Components/Apromore-Manager/src/main/java/org/apromore/cache/ehcache/TemporaryCacheService.java
```diff
@@ -0,0 +1,459 @@
...
+    return config;
+  }
+
+  public void setConfig(ConfigBean config) {
+    this.config = config;
+  }
+
+  private static final String APMLOG_CACHE_KEY_SUFFIX = "APMLog";
+
+  public String storeProcessLog(final Integer folderId, final String logName, XLog log,
```
franklma, line 105: Thank you for the time and effort taken to separate cache from manager package. Obviously, it was a bad idea to put the cache in there. 
Regarding this class, probably we should not move all the log related methods here. I prefer to decouple caching logic from these methods and put them back to EventLogService. Of course, I'm happy to do this later considering this is a temporary step.

franklma (approved):

karthik-balasubramanian (approved):

raboczi (approved):

Apromore-Core-Components/Apromore-Manager/src/main/java/org/apromore/cache/ehcache/TemporaryCacheService.java
nolantellis: Exactly. Thats the next. But I am aiming for db first. Hence the name temp 👍

PR #501, opened 2020-10-07
Remove Caching to new module

nolantellis:
Moved Caching to a seperate module. 
We will be able to now upgrade or work on caching as standalone and no need of manager.


brucenguyen1 (approved):

raboczi (approved):
One might quibble about locating Apromore-Cache directly under ApromoreCore rather than Apromore-Core-Components, but honestly I tend to prefer flatter source layouts so I'm not going to hold up the merge for it.  I am enthusiastic about this progress towards decomposing the manager!

PR #502, opened 2020-10-09
AP-1979 Support Unicode filenames

dreamind:


Apromore-Custom-Plugins/Process-Discoverer-Portal-Plugin/src/test/java/org/apromore/plugin/portal/processdiscoverer/common/ValidateFilenameTest.java
```diff
@@ -0,0 +1,89 @@
...
+                String filename = VALID_FILENAMES[i];
+                if (!Pattern.matches(pattern, filename)) {
+                    fail("Invalid name " + filename);
+                }
+            }
+            System.out.println("Testing valid filenames");
+        } catch (Exception e) {
+            fail("Exception occurred: " + e.getMessage());
+        }
+    }
```
brucenguyen1: I think this test aims to test whether PD will detect valid/invalid user input in the Input Dialog (org.apromore.plugin.portal.processdiscoverer.utils.InputDialog)? If so, can we relocate and rename it to do so making it easier to understand the purpose of this test? For example, it can be renamed as InputDialogTest, and the specific method to test here is InputDialog.okHandler, it can pass different values to txtValue, valuePatttern, etc. and check the expected value for labelError.

brucenguyen1: If it is not straightforward to instantiate ZK UI components in the unit test, the InputDialog needs certain refactoring to do to separate the logic out of the UI code for testabbility, making the code easier to maintain as well. For example, add a new class called InputChecker under the same package with a checkInput method, this method has basically the same input arguments as InputDiaglog.okHandler but with String types rather than UI components. Then, create a unit test named InputCheckerTest with similar code as ValidateFilenameTest. InputDialog.okHandler would call to the InputChecker class.

raboczi: Perhaps "ConstantsUnitTest" since it's just validating the correctness of the regexp VALID_FILENAME_REGEXP.

brucenguyen1: Unit tests if written should reflect the context of the system being tested, and so the naming of the test classes should carry that meaning. In this case, the code to be tested is mostly related to the pattern (it should also test the missing input though), but the naming of the test class is quite important for somebody to read the code.

dreamind: I have moved the filename test to a more general portal.common since it may be needed by portal components and to separate logic and UI. The ZK related test is probably done better via e2e/selenium tests.

brucenguyen1: Agreed, dreamind, just a minor comment: https://github.com/apromore/ApromoreCore/commit/8771d4e334d6f111f68e2037d3b7db44c6ae8a64#r43100604

brucenguyen1 (approved):

nolantellis (approved):

PR #503, opened 2020-10-09
Move db module out

nolantellis:
* This is moving the database out in a seperate module.
* I have commented 4 tests,which I will address before merging this
* I have commented event aspect which does some history service work. Will try to enable that. Else its not used anyways after discussion with simon and bruce.

* Had to clean up relationship between entites as they were not correct and effecting the code.


nolantellis:
Have Tested uploading logs, Creating user, Saving bpm models,creating new bpm models, Creating new versions, 
Creating folders, Repeating same for folders.

PR #505, opened 2020-10-12
Seperate DB into its own module

nolantellis:
* Moved Db in its own module.
* Corrected Test cases
* Activated AOP back.


nolantellis:
AOP doesnot eem to work. I am not uncommenting the historyAspect in the xml file, Will keep it there.

raboczi (approved):
I merged from ApromoreCore/development for compatibility with the latest ApromoreEE/development so that the AboutPlugin starts correctly, and did a few trivial whitespace restylings.  This builds and runs correctly as far as I've been able to test.

franklma (approved):

PR #507, opened 2020-10-12
Re-enabled pre-existing Checkstyle code. (AP-2409)

raboczi:
This PR is intended for commentary, not for merging any time soon.

Updated Checkstyle from v2 to v3.  Enabled validation on About-Portal-Plugin as a concrete example, showing that we can gradually enable style checks on a per-component basis using the checkstyle.skip property in the pom.xml.

The checkstyle.xml is mostly default; the rules Javadoc, DesignForExtension, and OperatorWrap are notable changes.  I've inserted a comment noting some of the rules from our current draft coding standard that can be captured by Checkstyle.  Each checkstyle rule has a justification at http://checkstyle.sf.net and these are noted in the comments within checkstyle.xml. 

The old (partially incompatible) checkstyle XML is retained as checkstyle-v2.xml but is not used; we may choose to translate some of the old rules to v3.  Because not all the old rules are compatible with v3, I've simply commented those out so that this old rule set does work in place of the new checkstyle.xml if so configured.

### DIFF
--- a/Apromore-Extras/Build-Tools/src/main/resources/checkstyle.xml
+++ b/Apromore-Extras/Build-Tools/src/main/resources/checkstyle.xml
@@ -20,7 +20,7 @@
     - some best practices
 
   Checkstyle is very configurable. Be sure to read the documentation at
-  http://checkstyle.sf.net (or in your downloaded distribution).
+  http://checkstyle.org (or in your downloaded distribution).
 
   Most Checks are configurable, be sure to consult the documentation.
 
@@ -55,7 +55,7 @@
 
     c. Each file should have at least a license header and comment explaining the purpose, usage and/or relationship with other classes (look at the JDK)
        -> Header (but we'll be using maven-license-plugin instead)
-       -> MissingJavadocType (see https://checkstyle.sourceforge.io/config_javadoc.html#MissingJavadocType)
+       -> MissingJavadocType (see https://checkstyle.org/config_javadoc.html#MissingJavadocType)
 
 -->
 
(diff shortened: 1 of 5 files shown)


nolantellis (approved):

Apromore-Extras/Build-Tools/src/main/resources/checkstyle.xml
```diff
@@ -1,256 +1,223 @@
...
+
+    <!-- Checks that a package-info.java file exists for each package.     -->
+    <!-- See http://checkstyle.sf.net/config_javadoc.html#JavadocPackage -->
+    <!-- <module name="JavadocPackage"/> -->
+
+    <!-- Checks whether files end with a new line.                        -->
+    <!-- See http://checkstyle.sf.net/config_misc.html#NewlineAtEndOfFile -->
+    <module name="NewlineAtEndOfFile"/>
+
+    <!-- Checks that property files contain the same keys.         -->
```
brucenguyen1: These conventions are much more than the coding standards. That's ok but suggest to add these conventions to the current coding standards. At least, a link to the checkstyle documentation site in the current standard would help anybody to have a good understanding than reading this checkstyle.xml. Btw, "http://checkstyle.sf.net" is obsolete, should be "https://checkstyle.org" to match with the maven checkstyle plugin site.

brucenguyen1: Sometimes the checkstyle plugin passed without detecting any deliberate violations. It's rare, but did happen, keep watching. Anyway, this is a good tool.

Apromore-Extras/Build-Tools/src/main/resources/checkstyle.xml
```diff
@@ -1,256 +1,223 @@
...
+    <module name="FileTabCharacter"/>
+
+    <!-- Miscellaneous other checks.                   -->
+    <!-- See http://checkstyle.sf.net/config_misc.html -->
+    <module name="RegexpSingleline">
+       <property name="format" value="\s+$"/>
+       <property name="minimum" value="0"/>
+       <property name="maximum" value="0"/>
+       <property name="message" value="Line has trailing spaces."/>
+    </module>
```
brucenguyen1: This rule prevents adding an empty line to format code blocks for logical flow and readability. The message "Line has trailing spaces" is misleading in this case. I think it aims to check non-empty lines with trailing spaces, not empty lines?

raboczi: You can add an empty line, but that line can't have whitespace to match the surrounding indentation; it has to be genuinely empty (i.e. just a newline).  Between this and disallowing tabs, you ensure that there's never any invisible whitespace.

brucenguyen1: On the other hand, consecutive empty lines should be checked but it seems there are no rules.

pom.xml
```diff
@@ -1468,20 +1468,26 @@
 						<artifactId>build-tools</artifactId>
 						<version>${build-tools.version}</version>
 					</dependency>
+					<dependency>
+						<groupId>com.puppycrawl.tools</groupId>
+						<artifactId>checkstyle</artifactId>
+						<version>8.36.2</version>
+					</dependency>
 				</dependencies>
```
brucenguyen1, line 1476: Can we put checkstyle.xml and other configuration files to top ApromoreCore module rather than putting them inside the build-tools module and then referring to them implicitly via classpath with a dependency on the build-tools module? Checkstyle is at a project-wide scope, just don't get why it must be inside build-tools which is for extra "tools".

Apromore-Extras/Build-Tools/src/main/resources/checkstyle.xml
brucenguyen1: "You can add an empty line, but that line can't have whitespace to match the surrounding indentation; it has to be genuinely empty (i.e. just a newline)": Tested and it doesn't pass that rule with the newline only.

brucenguyen1: Revised: my Eclipse automatically adds whitespaces for indentation, so press Enter to start a new line will add a number of white spaces and a newline character. It would need a little hassle to pass this rule.

pom.xml
raboczi: Putting checkstyle.xml at the top level would work -- it's how I initially did it, until realizing how Cameron had done it better.   Doing it through the build-tools module is better because it allows components to be built using only dependencies from the Maven repo (e.g. build-tools) rather than requiring the entire ApromoreCore source tree to be present.

raboczi:
Modified the trailing whitespace rule to only apply to lines with some non-whitespace.  This should be compatible with auto-indenting IDEs.  Normalized documentation links in the checkstyle.xml to use checkstyle.org rather than its aliases checkstyle.sf.net and checkstyle.sourceforge.io.

PR #508, opened 2020-10-12
Activated XLSX support

odaghagheleh:
In this commit, I have added JUnit tests for XLSX. I also fixed the Time zone issues we had with the tests. After the merge, we should able to support XLSX log import and also run all JUnit test successfully. 
The only task remaining for XLSX is providing support for different text encoding which I’m working on it right now.


raboczi (approved):
I see what you mean about the fix for the time zone issues in the tests not being ideal (i.e. just chopping off the zone field entirely).  For all the trouble these tests have caused us, I feel like they've been usefully reminding us that time zones ARE an issue when the software runs somewhere else.

Built and tested with a simple XSLX, so I'm satisfied this is okay to merge.

PR #509, opened 2020-10-13
AP-1160 Refactor date, time and duration handling

dreamind:


dreamind:
@brucenguyen1 I refactored some portion of the PD for date and time handling. There are lots of duplicate codes in logfilter, dash and apmlog that I have not touched, but this is just the first cut. I've made some changes that change the display the duration of a full time unit (e.g. 1 minute instead 60seconds), to make it consistent with dash and logfilter. Let me know if this is fine.

raboczi (approved):
This seems worthwhile for the sake of removing repeated code, so I approve its addition.

Perhaps storing the Constants fields as DateTimeFormatter instead of String would simplfy things?  They don't ever seem to be used without needing to be converted.  I feel like the "humanize" method could have a better name, maybe "formatMillisecondsSinceEpoch".

dreamind:
@raboczi Thanks, DateTimeFormatter is a good idea. The term humanize is quite a popular term in JS for such a purpose. It has been used in Python and Java as well (https://github.com/mfornos/humanize).

Apromore-Core-Components/Apromore-Manager/pom.xml
```diff
@@ -91,8 +91,10 @@
...
                             org.apromore.service.search,
-                            org.apromore.util
-                        </Export-Package>                        
+                            org.apromore.util,
+                            org.apromore.util.datetime
+                        </Export-Package>
+                        <JPA-PersistenceUnits>Apromore</JPA-PersistenceUnits>
                     </instructions>
                 </configuration>
             </plugin>
```
brucenguyen1: Agree this is a good refactoring to do. Can we put it in a separate module called apromore-commons as discussed before? The common miscellaneous utilities like date, time, stats, math, string, files, etc. are good candidates to be in a "commons" package (like apache-commons), this is the first and right time to do that. Apromore-Manager is too big, will have to be refactored and thus shouldn't be added more.

raboczi: The JPA-PersistenceUnits manifest entry at line Apromore-Manager/pom/xml:97 was removed as part of Nolan's recent factoring-out of the persistence into Apromore-Database.  I think its return is unintended.

Apromore-Core-Components/Apromore-Manager/src/main/java/org/apromore/util/datetime/DurationUnit.java
```diff
@@ -0,0 +1,100 @@
...
+ */
+
+package org.apromore.util.datetime;
+
+import java.util.Map;
+import java.util.HashMap;
+import java.time.temporal.ChronoUnit;
+
+public final class DurationUnit {
+
```
brucenguyen1: Please move the comment "/** Duration unit...." to above the class DurationUnit, rather than for the package. The package will be repeated for each class file, the comment is for each class, not for the package.

Apromore-Core-Components/Apromore-Manager/src/main/java/org/apromore/util/datetime/DurationUtil.java
```diff
@@ -0,0 +1,124 @@
...
+import java.text.DecimalFormat;
+import java.time.temporal.ChronoUnit;
+
+import org.apromore.util.datetime.Constants;
+
+/**
+ * Various duration utility functions
+ */
+public final class DurationUtil {
+
```
brucenguyen1: Same as above, please move the comment above the package to above the class definition.

brucenguyen1 (commented):
This is a good refactoring to do. As it is the first cut while collecting date time usages from different modules, suggest the focus for this time would be: (1) place these utilities in a separate module named apromore-commons instead of Apromore-Manager, (2) collecting date time usages from different modules first in order to form a consistent/meaningful API for these utilities. It looks like the API will change a lot due to naming issues, leading to various changes to the unit tests and other dependent modules.

Apromore-Core-Components/Apromore-Manager/pom.xml
```diff
@@ -91,8 +91,10 @@
                             org.apromore.service.helper,
                             org.apromore.service.model,
                             org.apromore.service.search,
-                            org.apromore.util
-                        </Export-Package>                        
+                            org.apromore.util,
+                            org.apromore.util.datetime
+                        </Export-Package>
+                        <JPA-PersistenceUnits>Apromore</JPA-PersistenceUnits>
```
nolantellis: Remove the JPA stuff from here. Its in a different module.

Apromore-Core-Components/Apromore-Manager/src/main/java/org/apromore/util/datetime/DateTimeUtil.java
```diff
@@ -0,0 +1,130 @@
...
+import java.time.format.DateTimeFormatter;
+
+import org.apromore.util.datetime.Constants;
+
+/**
+ * Various date and time utility functions
+ */
+public final class DateTimeUtil {
+
+    public static LocalDateTime parse(String lastUpdate, DateTimeFormatter formatter) {
```
nolantellis: rename lastUpdate to a timestamp name

Apromore-Core-Components/Apromore-Manager/src/main/java/org/apromore/util/datetime/DateTimeUtil.java
```diff
@@ -0,0 +1,130 @@
...
+        LocalDateTime dateTime;
+        try {
+            dateTime = LocalDateTime.parse(lastUpdate, formatter);
+        } catch (Exception e) {
+            dateTime = null;
```
nolantellis: Maybe some logging her emight help to understand what comes here besides the relevent format.

Apromore-Core-Components/Apromore-Manager/src/main/java/org/apromore/util/datetime/DateTimeUtil.java
```diff
@@ -0,0 +1,130 @@
...
+
+    public static String format(long milliseconds, String format){
+        return format(toZonedDateTime(milliseconds), format);
+    }
+
+    public static String humanize(long milliseconds){
+        return format(milliseconds, Constants.DATE_TIME_FORMAT_HUMANIZED);
+    }
+
+}
```
nolantellis: Let me try something here. Before you merge in dev. Contact me when ur done.

Apromore-Core-Components/Apromore-Manager/src/main/java/org/apromore/util/datetime/DurationUnit.java
```diff
@@ -0,0 +1,100 @@
...
+    private static Map<ChronoUnit, String> shortLabelPluralMap = new HashMap<ChronoUnit, String>() {
```
nolantellis: Enum is the best to beused here.

Apromore-Core-Components/Apromore-Manager/src/test/java/org/apromore/util/datetime/DateTimeUtilUnitTest.java
```diff
@@ -0,0 +1,79 @@
...
+import java.time.LocalDateTime;
+import java.time.ZoneId;
+import java.time.ZonedDateTime;
+
+import org.apromore.util.datetime.Constants;
+import org.apromore.util.datetime.DateTimeUtil;
+
+import org.junit.Test;
+
+public class DateTimeUtilUnitTest {
```
nolantellis: Have a look at parameterizedTest in junit.

nolantellis (changes requested):

Apromore-Core-Components/Apromore-Manager/src/main/java/org/apromore/util/datetime/DurationUnit.java
dreamind: Initially plan to "extends" ChronoUnit enum, but it looks too complicated. I end up keeping the ChronoUnit enum.

dreamind:
@nolantellis Cool with all the changes. Could you approve please?

brucenguyen1 (approved):

PR #513, opened 2020-10-14
Ap 2304 Backend service for sharing artifact (Log, Process, Folder)

franklma:
Add sharing APIs in WorkspaceService, which basically a mapper between ACL access type and flags in group_* tables.


Apromore-Core-Components/Apromore-Manager/src/main/java/org/apromore/service/impl/UserMetadataServiceImpl.java
```diff
@@ -141,6 +142,42 @@ public void saveUserMetadata(String userMetadataContent, UserMetadataTypeEnum us
...
+    private  Set<GroupUsermetadata> findByLogAndGroup(Integer logId, String groupRowGuid) {
+
+        Group group = groupRepo.findByRowGuid(groupRowGuid);
+        Set<GroupUsermetadata> result = new HashSet<>();
+
+        // All the user metadata that linked to this log
+        Set<UsermetadataLog> usermetadataLogSet =
+                new HashSet<>(usermetadataLogRepo.findByLog(logRepo.findUniqueByID(logId)));
+
+        if (usermetadataLogSet.size() != 0) {
```
nolantellis: if loop not required.

Apromore-Core-Components/Apromore-Manager/src/main/java/org/apromore/service/impl/UserMetadataServiceImpl.java
```diff
@@ -141,6 +142,42 @@ public void saveUserMetadata(String userMetadataContent, UserMetadataTypeEnum us
...
+
+            for (UsermetadataLog usermetadataLog : usermetadataLogSet) {
+                Usermetadata u = usermetadataLog.getUsermetadata();
```
nolantellis, line 165: Not important now, but a quick way to do this is get all the Ids in 1 db quiery, and then use in query to get all list of groupmeta data , and then so the needful.

Apromore-Core-Components/Apromore-Manager/src/main/java/org/apromore/service/impl/UserMetadataServiceImpl.java
```diff
@@ -182,28 +219,18 @@ public void saveUserMetadataPermissions(Integer logId, String groupRowGuid, bool
...
-                List<GroupUsermetadata> groupUsermetadataList = groupUsermetadataRepo.findByUsermetadataId(u.getId());
+        for (GroupUsermetadata g : groupUsermetadataSet) {
 
-                // Remove permissions assigned to specified group
-                for (GroupUsermetadata g : groupUsermetadataList) {
-                    if (g.getGroup().equals(group)) {
-                        groupUsermetadataRepo.delete(g);
-                    }
-                }
+            if(g.getUsermetadata().getGroupUserMetadata().size() == 1) { // if this metadata will be an orphan
```
nolantellis, line 230: Not sure, but you may get nullpointer over here if getUserMetadata is null

Apromore-Core-Components/Apromore-Manager/src/main/java/org/apromore/service/impl/WorkspaceServiceImpl.java
```diff
@@ -695,26 +697,125 @@ private void removeGroupLog(Group group, Log log) {
...
-//        if (groupLog != null) {
-//            groupLogRepo.delete(groupLog);
-//        }
-//    }
+    public Map<Group, AccessType> getLogACL(Integer logId) {
+
+        Map<Group, AccessType> groupAccessTypeMap = new HashMap<>();
+
+        for (GroupLog g : getGroupLogs(logId)) {
+            if (g.getHasRead() && g.getHasWrite() && g.getHasOwnership()) {
```
nolantellis: you can group. (g.getHasRead() && g.getHasWrite() && g.getHasOwnership()) inside GroupLog with helper method isOwner() or anyother name. and make this 1 line ternary operator

Apromore-Core-Components/Apromore-Manager/src/main/java/org/apromore/service/impl/WorkspaceServiceImpl.java
```diff
@@ -695,26 +697,125 @@ private void removeGroupLog(Group group, Log log) {
...
+
+        return groupAccessTypeMap;
+    }
+
+    public Map<Group, AccessType> getProcessACL(Integer processId) {
+
+        Map<Group, AccessType> groupAccessTypeMap = new HashMap<>();
+
+        for (GroupProcess g : getGroupProcesses(processId)) {
+            if (g.getHasRead() && g.getHasWrite() && g.getHasOwnership()) {
```
nolantellis: Repeated code

nolantellis (changes requested):
Things look alright. If you want to make the changes suggested , let me know, I may help in some changes. Else to get this work done quickly ,your free to merge.

dreamind (approved):

Apromore-Core-Components/Apromore-Manager/src/main/java/org/apromore/util/AccessType.java
```diff
@@ -0,0 +1,50 @@
...
+    private final Integer flagId;
+
+    AccessType(Integer flagId) {
+        this.flagId = flagId;
+    }
+
+    public Integer getFlogId() {
+        return flagId;
+    }
+}
```
brucenguyen1: Misspelling: getFlagId.

Apromore-Core-Components/Apromore-Manager/src/main/java/org/apromore/util/AccessType.java
```diff
@@ -0,0 +1,50 @@
...
+    NONE(0),
+    VIEWER(1),
+    EDITOR(2),
+    OWNER(3);
+
+    private final Integer flagId;
+
+    AccessType(Integer flagId) {
+        this.flagId = flagId;
+    }
```
brucenguyen1, line 45: Missing access modifier.

brucenguyen1: What happens if flagId is an arbitrary integer, e.g. 100?

Apromore-Core-Components/Apromore-Manager/src/main/java/org/apromore/service/impl/WorkspaceServiceImpl.java
franklma: Are there ways to reduce repeating code here other than abstract comparing and use ternary operator? Thanks

Apromore-Core-Components/Apromore-Manager/src/main/java/org/apromore/service/WorkspaceService.java
```diff
@@ -174,4 +173,75 @@
...
+    Map<Group, AccessType> getProcessACL(Integer processId);
+
+    /**
+     * Get list of Group and AccessType pair of Log
+     * @param folderId Folder ID
+     * @return list of Group and AccessType pair
+     */
+    Map<Group, AccessType> getFolderACL(Integer folderId);
+
+    /**
```
brucenguyen1: These XXXPermissions and XXXACL should be put in a separate class which is responsible for security/access right management. The WorkspaceService is responsible for managing the workspace such as create/remove/move folder, get list of subfolders, breadcrumbs, etc. The general principle is "one class one responsiblity". Access rights/permissions belong to a different type of responsiblity. The reason we need to care about responsibility in class design is because the class would be modified when its responsiblity changes, and thus, keep responsiblities in separate classes would help keep the code clean and minimal impact on the code when changes happen (always happen). Note that there is already a service class called SecurityService with a mix of authentication (username, password) and authorization (roles) - it's not clean, suggest to move XXXPermissions and XXXACL to a class called something like AuthorizationService, PermissionService, or AccessRightService.

Apromore-Core-Components/Apromore-Manager/src/main/java/org/apromore/util/AccessType.java
brucenguyen1: flagId can be an internal variable name, but the method name getFlagId() doesn't show its meaning. It can be given a better name according its intention here: e.g. getAccessTypeCode? I would suggest to rename the flagId variable to be a better name like accessTypeCode.

Apromore-Core-Components/Apromore-Manager/src/test/java/org/apromore/service/impl/WorkspaceServiceImplTest.java
```diff
@@ -362,5 +353,122 @@ public void testMoveFolder() throws Exception {
...
+        Group group2 = createGroup(2, Group.Type.GROUP);
+        Group group3 = createGroup(3, Group.Type.USER);
+        Group group4 = createGroup(4, Group.Type.USER);
+
+        Role role = createRole(createSet(createPermission()));
+        User user = createUser("userName1", group1, createSet(group1), createSet(role));
+
+        Workspace wp = createWorkspace(user);
+        Log log = createLog(user, createFolder("testFolder", null, wp));
+
```
brucenguyen1: Unit test methods should be named as testXXX, e.g. testGetLogACL, or test_getLogACL, testGetProcessACL, etc. This is to avoid miscommunication when referring to a test method: not to mislead it with the method under test.

Apromore-Core-Components/Apromore-Manager/src/main/java/org/apromore/service/impl/WorkspaceServiceImpl.java
brucenguyen1: This method should be renamed as: getLogAccessType() as shown from its implementation, the return is AccessType. ACL is Access Control List which is not at this level. Same for other method: getFolderACL, getXXXACL.

Apromore-Core-Components/Apromore-Manager/src/main/java/org/apromore/service/impl/UserMetadataServiceImpl.java
brucenguyen1: What happens if an arbitrary groupRowGuid value is passed to the findByLogAndGroup() method and a null is returned at line 154? This kind of corner case can be checked in other places and would be great to be in unit tests to be solid.

Apromore-Core-Components/Apromore-Manager/src/main/java/org/apromore/service/impl/WorkspaceServiceImpl.java
brucenguyen1: This block of code: {if...then...else...} with hasRead, hasWrite, hasOwnership need a higher level of abstraction to minimize various miscellaneous changes to source code in many places like this if these any rules to hasRead, hasWrite, etc. change (which always happens). From the repeated code block, higher abstraction methods are needed to hide these details: e.g. checkHasEditorAccessType(), checkHasViewerAccessType, checkHasOwnerAccessType. Again, these authorization checks should be responsibilty of a separate class.

Apromore-Core-Components/Apromore-Manager/src/main/java/org/apromore/service/WorkspaceService.java
franklma: Yes. AuthorizationService class sounds a good place to contains access control methods, and we can move all the authorization (roles) methods from SecurityService to this class in the next step.

Apromore-Core-Components/Apromore-Manager/src/main/java/org/apromore/service/impl/UserMetadataServiceImpl.java
franklma: Will add unit tests to cover these corner cases.

Apromore-Core-Components/Apromore-Manager/src/main/java/org/apromore/service/impl/WorkspaceServiceImpl.java
brucenguyen1: Alternatively, you can create just one method called checkPermissionByAccessType(AccessType accessType, Permission permission). Note that a new class called Permission is introduced here to wrap the details of hasRead, hasOwner, hasXXX. It makes the code much cleaner. Btw, as you're adding code relating to these permissions, I suggest we need a little refactoring here: adding the Permission class and replacing all references to details of hasXXX to this Permission class.

brucenguyen1 (commented):
I have a number of comments. Some are related to a little bit refactoring which is needed to improve code while the new code is added.

Apromore-Core-Components/Apromore-Manager/src/main/java/org/apromore/service/impl/WorkspaceServiceImpl.java
franklma: Yes. getLogAccessType is definitely more accurate.

Apromore-Core-Components/Apromore-Manager/src/test/java/org/apromore/service/impl/WorkspaceServiceImplTest.java
franklma: RIght. They were auto-generated by IntelliJ.

Apromore-Core-Components/Apromore-Manager/src/main/java/org/apromore/service/impl/WorkspaceServiceImpl.java
```diff
@@ -695,26 +697,125 @@ private void removeGroupLog(Group group, Log log) {
...
+
+    public void deleteProcessACL(Integer processId, String groupRowGuid) {
+
+        removeProcessPermissions(processId, groupRowGuid);
+    }
+
+    public void deleteFolderACL(Integer folderId, String groupRowGuid) {
+
+        removeFolderPermissions(folderId, groupRowGuid);
+    }
```
brucenguyen1: Relating to one comment above, these methods deleteXXXACL with a one-to-one mapping to removeXXXPermissions show that ACL = Permission. That's why the above comment suggests using a better name for getXXXACL (they should be getXXXAccessType instead), and ACL corresponds to Permission, so suggest using one consistent name in source code: either ACL or Permission, not both. Once consistent notion and name have been determined, these one-to-one mapping methods will not be needed any more.

brucenguyen1 (approved):
As discussed, refactoring will need to do later in terms of method naming (mixed up feature access with resource access), method signature, and separation of code  to classes.

PR #517, opened 2020-10-15
Ap 2419 Add preConditions in LiquidBase changeSet to cover all scenarios

franklma:


Apromore-Database/src/main/resources/db/migration/changeLog-Schema.yaml
```diff
@@ -464,7 +464,16 @@ databaseChangeLog:
...
+     preConditions:
+       - onFail: MARK_RAN
+       - not:
+           - or:
+               - foreignKeyConstraintExists:
+                   foreignKeyName: fk_user_group
+               - indexExists:
+                   indexName: idx_search_history
+               - indexExists:
+                   indexName: group_row_guid_UNIQUE
```
brucenguyen1: As foreign key name and index name are unique within a table but not across tables, adding "tableName: <tablename>" under foreignKeyConstraintExists and indexExists would be more accurate.  But it might not really necessary if a consistent naming convention is used for foreign key names and index names. It would be a problem in case of lacking a consistent naming convention.

group_row_guid_UNIQUE: it should have "idx" prefix to be consistent index name, like "idx_search_history".

brucenguyen1 (approved):
Tested and it works in both conditions: when a table/index/constraint exist or not exist. A consistent naming convention is quite important to follow.

Apromore-Database/src/main/resources/db/migration/changeLog-Schema.yaml
franklma: Surprisingly, FK and indexExists conditions don't support table attribute, according to the official document.
https://docs.liquibase.com/concepts/advanced/preconditions.html

brucenguyen1: You can try it, Frank. There are examples in here: https://github.com/liquibase/liquibase/blob/master/liquibase-integration-tests/src/test/resources/changelogs/yaml/common.tests.changelog.yaml. For foreign key: no need because it must be unique in a database.

franklma: Yes. I tested and working fine. Looks like the preConditions part is poorly documented.

PR #519, opened 2020-10-19
AP-2432 Add additional quick filter shortcut in PD

dreamind:


chiichang:
intermediate stage sufficient for testing the shortcuts

PR #520, opened 2020-10-19
Feature/ap 2416 hadoop on windows

raboczi:
There are two separate changes involved here:

(1) Precompiled libraries for Windows 10 are included in Supplements/hadoop-3.2.1.  These allow access by the Hadoop libraries to the local filesystem when running on Windows.  The **hadoop.home.dir** system variable is set both for the running server and for CSV Importer logic's JUnit tests.  Note that the wrapped libraries in parquet-osgi are Hadoop version 3.3.0, whereas the latest precompiled Windows binaries available from https://github.com/cdarlint/winutils are the slightly earlier 3.2.1.  The test suite for CSV Importer logic still doesn't pass on Windows, but now it's for reasons unrelated to Hadoop (failures rather than errors).  This addresses https://apromore.atlassian.net/browse/AP-2416.

(2) Incidental refactoring for the CSV Importer logic's test suite, collecting the code which reads test files out of the bundle classpath: **TestUtilities.resourceToString**.  This was motivated by trying to solve the test failures, under the assumption that it had something to do with different line endings of Windows.  Actually fixing it has been deferred as a separate task ([AP-2447](https://apromore.atlassian.net/browse/AP-2447)).


odaghagheleh:
Thanks, Simon, I will pull and test the code and make sure all the test passing. In the meantime It will be good if @brucenguyen1 can test it on Windows and make sure Hadoop is working (junit test passes)

odaghagheleh:
All test passed and could import Parquet successfully

brucenguyen1 (commented):
No test error "HADOOP_HOME and hadoop.home.dir are unset", but unit tests still haven't passed, not recommended for merging into development until all tests can pass.

PR #523, opened 2020-10-20
Bug/dbweaving on windows/ap 2437

nolantellis:
* Solve static weaving exception for windows.
* Small bug fix in datetime parsing


raboczi (changes requested):
Could you switch the destination from "master" to "development"?

raboczi (approved):
Confirmed that it now builds (and runs) on Windows.

PR #524, opened 2020-10-20
AP-2120 Refine merge model interface

dreamind:
- Some renames for common to follow apache.commons
- Refactor ItemUtil into FilenameUtils and add mergeNames (follow apache.commons.io.FilenameUtils)
- Refine merge model plugins


Apromore-Commons/pom.xml
```diff
@@ -16,18 +16,17 @@
   <packaging>jar</packaging>
   <name>Apromore-Commons</name>
 
+  <properties>
```
nolantellis, line 19: This is good, There are many new features. If you get the time and bandwitdth please do this for the core project. Not sure how much backward compatible it will be though.

Apromore-Commons/pom.xml
```diff
@@ -36,12 +35,54 @@
...
+    <dependency>
+      <groupId>org.junit.jupiter</groupId>
+      <artifactId>junit-jupiter-engine</artifactId>
+      <version>${junit.jupiter.version}</version>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.junit.platform</groupId>
+      <artifactId>junit-platform-launcher</artifactId>
+      <version>1.4.2</version>
```
nolantellis, line 59: If u look at  junit 5 docs, I dont think we need so many imports, not sure but may be cross check again.

nolantellis (approved):

Apromore-Commons/src/main/java/org/apromore/commons/io/FilenameUtils.java
```diff
@@ -0,0 +1,122 @@
...
+import java.util.Collections;
+import java.util.List;
+/**
+ * Various filename utility functions
+ */
+public final class FilenameUtils {
+
+    private static final String DELIMITERS = "\\._\\+\\- ";
+    private static final String MERGED = "_merged";
+    private static final int MAX_LENGTH = Constants.VALID_FILENAME_MAX_LENGTH - MERGED.length();
```
brucenguyen1: Suggest renaming the package name from "org.apromore.commons.io" to "org.apromore.commons.item", and the class name from FilenameUtils to ItemNameUtils. Apromore has been using "item" for process models, logs (e.g. SummaryItemRenderer in Apromore-Portal) and "Item" seems to be OK and a clean name to be used for these things, but "File" is not OK as it would be confusing with the "File" terminology used for computer files, directories as described in Apache-Commons' FilenameUtils javadocs (https://commons.apache.org/proper/commons-io/javadocs/api-release/org/apache/commons/io/FilenameUtils.html), and IO is of course a standard term used for computer files, streams, etc. as described in apache.commons.io package javadocs (http://commons.apache.org/proper/commons-io/apidocs/index.html?org/apache/commons/io/package-summary.html). We should not mix up these two notions: "Item" in Apromore and the standard term "File" for computer files.

brucenguyen1 (approved):

PR #525, opened 2020-10-20
APIs to get dependent artifact of user metadata

franklma:
1. APIs to get dependent artifact (Log, Process) of user metadata
2. Refactoring of UserMetadataService
3. Unit tests


dreamind:
@frank-ma-leonardo Do we have a method to get the usermetadata AccessType (e.g. getUserMetadataAccessType)?

Apromore-Core-Components/Apromore-Manager/src/main/java/org/apromore/service/impl/UserMetadataServiceImpl.java
```diff
@@ -281,65 +280,94 @@ public void deleteUserMetadata(Integer usermetadataId, String username) throws U
...
         }
+        return umSet;
+    }
+
+    @Override
+    public Set<Usermetadata> getUserMetadataByLogs(List<Integer> logIds, UserMetadataTypeEnum userMetadataTypeEnum) {
+
+        if (null == logIds || logIds.size() == 0) {
+            return null;
+        }
```
brucenguyen1: Return an empty set would help avoid NPE at the calling side and make code less "if else" block.

Apromore-Core-Components/Apromore-Manager/src/main/java/org/apromore/service/impl/UserMetadataServiceImpl.java
```diff
@@ -281,65 +280,94 @@ public void deleteUserMetadata(Integer usermetadataId, String username) throws U
...
+        Set<Usermetadata> umSetLinkedToUser = getUserMetadataByUser(username, userMetadataTypeEnum);
+        Set<Usermetadata> umSetLinkedToLogs = getUserMetadataByLogs(logIds, userMetadataTypeEnum);
+
         // Find intersection of user metadata lists that get from 2 linked tables (log, group)
-        result.retainAll(usermetadataList1);
+        umSetLinkedToUser.retainAll(umSetLinkedToLogs);
 
-        return result.size() > 0 ? result : null;
+        return umSetLinkedToUser.size() > 0 ? umSetLinkedToUser : null;
     }
```
brucenguyen1: Return an empty set would help avoid NPE at the calling side and make code less "if else" block.

Apromore-Core-Components/Apromore-Manager/src/test/java/org/apromore/service/impl/UserMetadataBuilder.java
```diff
@@ -0,0 +1,34 @@
...
+    }
+
+    public UserMetadataBuilder withUsermetadataType(Integer id, String type){
+        usermetadataType = new UsermetadataType();
+        usermetadataType.setIsValid(true);
+        usermetadataType.setType(type);
+        usermetadataType.setVersion(1);
+        usermetadataType.setId(id);
+        return this;
+    }
```
brucenguyen1, line 29: I don't get this method: usermetadataType would be useless.

Apromore-Core-Components/Apromore-Manager/src/test/java/org/apromore/service/impl/UserMetadataBuilder.java
```diff
@@ -0,0 +1,34 @@
...
+
+public class UserMetadataBuilder {
+
+    Group group;
+    UsermetadataType usermetadataType;
+    Usermetadata usermetadata;
+
+    Set<GroupUsermetadata> groupUsermetadataSet = new HashSet<>();
+    Set<UsermetadataLog> usermetadataLogSet = new HashSet<>();
+    Set<UsermetadataProcess> usermetadataProcessSet = new HashSet<>();
```
brucenguyen1: Need access modifier: private, public?

Apromore-Core-Components/Apromore-Manager/src/test/java/org/apromore/service/impl/UserMetadataBuilder.java
```diff
@@ -0,0 +1,34 @@
...
+        usermetadataType.setType(type);
+        usermetadataType.setVersion(1);
+        usermetadataType.setId(id);
+        return this;
+    }
+
+    public UsermetadataType buildUsermetadataType() {
+        return usermetadataType;
+    }
+}
```
brucenguyen1: Don't get these two methods. If you need to create a UsermetadataType from an "id" and "type", it can be done with buildUsermetadataType(Integer id, String type), why you need two calls: withUsermetadataType(Integer id, String type) and then buildUsermetadataType() with no parameters? In addition, buildUsermetadataType can be a static method if this class is only to create a UsermetadataType from an id and type param and so the caller won't need to instantiate any UserMetadataBuilder object, just call UserMetadataBuilder.buildUsermetadataType.

Apromore-Database/src/main/java/org/apromore/dao/model/Usermetadata.java
```diff
@@ -79,6 +79,10 @@
      * Set of UserMetadataLog from linked table
      */
     private Set<UsermetadataLog> usermetadataLogSet = new HashSet<>();
+    /**
+     * Set of UserMetadataLog from linked table
+     */
+    private Set<UsermetadataProcess> usermetadataProcessSet = new HashSet<>();
```
brucenguyen1, line 86: Can we keep naming consistent across all classes and methods regarding "Usermetadata" or "UserMetadata"?
Some examples: Usermetadata, UsermetadataLog, UsermetadataProcess, UserMetadataTypeEnum, UserMetadataBuilder, UsermetadataType, getUserMetadataByLog, getUserMetadata, GroupUsermetadata.

Apromore-Core-Components/Apromore-Manager/src/main/java/org/apromore/service/impl/UserMetadataServiceImpl.java
```diff
@@ -281,65 +280,94 @@ public void deleteUserMetadata(Integer usermetadataId, String username) throws U
...
-    public Set<Usermetadata> getUserMetadata(String username, List<Integer> logIds,
-                                             UserMetadataTypeEnum userMetadataTypeEnum) throws UserNotFoundException {
-
-        if (null == logIds || logIds.size() == 0) {
-            return null;
-        }
+    public Set<Usermetadata> getUserMetadataByUser(String username, UserMetadataTypeEnum userMetadataTypeEnum) throws UserNotFoundException {
 
         User user = userSrv.findUserByLogin(username);
         assert user != null;
```
brucenguyen1: This line is redundant because findUserByLogin will throw a UserNotFoundException: assert user != null;

Apromore-Core-Components/Apromore-Manager/src/main/java/org/apromore/service/impl/UserMetadataServiceImpl.java
```diff
@@ -281,65 +280,94 @@ public void deleteUserMetadata(Integer usermetadataId, String username) throws U
...
-        Set<Usermetadata> usermetadataList1 = new HashSet<>();
-        for (GroupUsermetadata groupUsermetadata : groupUsermetadataSet) {
-            Usermetadata u = groupUsermetadata.getUsermetadata();
-            if (u.getUsermetadataType().getId().equals(userMetadataTypeEnum.getUserMetadataTypeId()) && u.getIsValid()) {
-                usermetadataList1.add(u);
+            List<GroupUsermetadata> guList = groupUsermetadataRepo.findByGroup(group);
+
+            for (GroupUsermetadata gu : guList) {
+                Usermetadata u = gu.getUsermetadata();
+                if (u.getUsermetadataType().getId().equals(userMetadataTypeEnum.getUserMetadataTypeId()) && u.getIsValid()) {
```
brucenguyen1, line 296: It would help a little bit to read code if this method is shorter: userMetadataTypeEnum.getUserMetadataTypeId() -> userMetadataTypeEnum.getCode().

Apromore-Core-Components/Apromore-Manager/src/main/java/org/apromore/service/UserMetadataService.java
```diff
@@ -58,6 +60,15 @@ void saveUserMetadataLinkedToOneLog(String userMetadataContent, UserMetadataType
...
                                         Integer logId) throws UserNotFoundException;
 
+    /**
+     * Save a user metadata which is not linked to log. So it can not be shared to other users at stage one.
+     *
+     * @param content Content of user metadata
+     * @param username username
+     * @throws UserNotFoundException Can't find a user with specified username
+     */
+    void saveUserMetadataWithoutLog(String content, UserMetadataTypeEnum userMetadataTypeEnum, String username) throws UserNotFoundException;
```
brucenguyen1, line 69: Keeping both UserMetadataType (values stored in database) and UserMetadataTypeEnum (values hard-coded) would create issues later if the hard-coded values in UserMetadataTypeEnum are inadvertently changed and out of sync with the values stored for UserMetadataType in the database. We should use consistently UserMetadataType in everywhere in the code if it needs to be used, no need another duplication like UserMetadataTypeEnum.

brucenguyen1: Suggest to replace UserMetadataTypeEnum param used in many methods with userMetadataTypeId (integer type).

Apromore-Database/src/main/java/org/apromore/dao/model/Usermetadata.java
franklma: It's a compromise due to JPA's naming convention.

Apromore-Core-Components/Apromore-Manager/src/main/java/org/apromore/service/UserMetadataService.java
franklma: Created a separate card for this refactoring.
https://apromore.atlassian.net/browse/AP-2454

frank-ma-leonardo:
> @frank-ma-leonardo Do we have a method to get the usermetadata AccessType (e.g. getUserMetadataAccessType)?

AccessType getUserMetadataAccessType(Group group, Usermetadata usermetadata);

dreamind (approved):

brucenguyen1 (approved):

franklma:
Created a separate card for naming issue:
https://apromore.atlassian.net/browse/AP-2458

PR #527, opened 2020-10-21
Bugfix/ap 2447 csvimporter tests on windows

raboczi:
These changes to the CSV Importer logic unit tests pass on both unix and Windows 10.  Most of the tests that originally failed under Windows only needed their expected test result documents to have any CR-LF (Windows style) line endings normalized to LF (unix style).  The test for CP-1255 (hebrew) additionally required the charset of the test document to be specified during String conversion.


raboczi:
I've run out of ideas for duplicating the failures that Bruce is seeing, and will probably have to hand it over to him to debug on his system as time allows.  The following command line was resulting in 58 successful tests for me:
Screen Shot 2020-10-27 at 14 48 57

git clone https://github.com/apromore/ApromoreCore.git -b bugfix/AP-2447-csvimporter-tests-on-windows
cd ApromoreCore
mvn test -pl :csvimporter-logic

PR #534, opened 2020-10-28
* Corrected documentation regarding database.

nolantellis:
* Corrected code to include new liquibase user.
* corrected documentation

### DIFF
--- a/site.properties
+++ b/site.properties
@@ -90,7 +90,7 @@ ldap.lastNameAttribute  = sn
 # Database and JPA (H2)
 
 jdbc.driver   = org.h2.Driver
-jdbc.url      = jdbc:h2:../Manager-Repository;DB_CLOSE_DELAY=-1;DATABASE_TO_UPPER=false;
+jdbc.url      = jdbc:h2:../Manager-Repository;DB_CLOSE_DELAY=-1;
 jdbc.username = apromore
 jdbc.password = MAcri
 
@@ -98,8 +98,8 @@ jpa.database         = H2
 jpa.databasePlatform = org.eclipse.persistence.platform.database.H2Platform
 jpa.showSql          = false
 jpa.generateDDL      = false
-
-# Liquibase User H2
+#
+## Liquibase User H2
 liquibase.jdbc.username = apromore
 liquibase.jdbc.password = MAcri
 
--- a/site.properties
+++ b/site.properties
@@ -107,7 +107,7 @@ liquibase.jdbc.password = MAcri
 
 #jdbc.driver   = com.mysql.jdbc.Driver
 #jdbc.url      = jdbc:mysql://localhost:3306/apromore?createDatabaseIfNotExist=true&autoReconnect=true&allowMultiQueries=true&rewriteBatchedStatements=true&characterEncoding=utf-8
-#jdbc.username = root
+#jdbc.username = apromore
 #jdbc.password = MAcri
 #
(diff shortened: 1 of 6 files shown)


site.properties
```diff
@@ -89,29 +89,35 @@ ldap.lastNameAttribute  = sn
...
-jdbc.password = MAcri
+#jdbc.driver   = com.mysql.jdbc.Driver
+#jdbc.url      = jdbc:mysql://localhost:3306/apromore?createDatabaseIfNotExist=true&autoReconnect=true&allowMultiQueries=true&rewriteBatchedStatements=true&characterEncoding=utf-8
+#jdbc.username = root
+#jdbc.password = MAcri
+#
+#jpa.database         = MYSQL
+#jpa.databasePlatform = org.eclipse.persistence.platform.database.MySQLPlatform
+#jpa.showSql          = false
+#jpa.generateDDL      = false
```
brucenguyen1: Why don't we use the same 'apromore' user like H2 instead of 'root'?

site.properties
```diff
@@ -89,29 +89,35 @@ ldap.lastNameAttribute  = sn
...
 
-jpa.database         = MYSQL
-jpa.databasePlatform = org.eclipse.persistence.platform.database.MySQLPlatform
-jpa.showSql          = false
-jpa.generateDDL      = false
+## Liquibase User Mysql
+#liquibase.jdbc.username = db_user
+#liquibase.jdbc.password = 7fHJV41fpJ
```
brucenguyen1: '7fHJV41fpJ': is it an encrypted value of the password or is it the password? If it is the password, then it should be encrypted as it can be seen easily in this file (it is full privileges).

site.properties
brucenguyen1: Can it is named "liquibase_user" instead of "db_user" (it is full privileges)? The name will be shown in the database management tool for information as well.

nolantellis: I am not sure how it became root. Changeing that back to apromore. It was a mistake.

db_user is just a name. And looking at the Docs of mysql about the privileges, All privileges is the right 1. (i.e ALL privileges does not include create user.)

site.properties
nolantellis: Its just a random passowrd. encryption is a different logic which we can consider later. Its not in this scope

brucenguyen1: Ok if it's a low security risk for showing password of an "All privileges" user in a text file. Any justification?

site.properties
nolantellis: Corrected to liquibase_usr

site.properties
nolantellis: The use is excactly the same as apromore user at this time.
All previleges is only for apromore db, and as per the mysql5.7 doc All privilege is DML and DDL and not USER management.

I Agree this is not the right way, but thats another scope to consider not only for liquibase but also for apromore user.

brucenguyen1: https://dev.mysql.com/doc/refman/5.7/en/grant.html (Table 13.8, the "Level" information). There's a risk of losing data due to password is revealed in plain text file for 'apromore' and 'liquibase_user', at the same risk level for both users. This risk should be addressed in the future if possible.

brucenguyen1 (approved):

raboczi (approved):
I'm satisfied that it starts properly with zero configuration under H2, and that it starts properly under MySQL given the setup described in README.md.

PR #535, opened 2020-10-28
Integrate Simulator v0.1-Save Simulation Parameters

stdevi:
Closes AP-1874

Epic link: https://apromore.atlassian.net/l/c/B5Q8EQct


raboczi:
The show-stopping problem when I build this branch is that the BPMN editor's "Save" and "Save As" toolbar buttons stop working.

The simulation parameter panel works nicely, although (like the rest of the system) it's a little awkward to use in dark mode on macOS.  (Not a problem that needs addressing any time soon.)

PR #536, opened 2020-10-30
Add license preamble

franklma:


raboczi (approved):
This is clearly not going to break anything at runtime.  :)

PR #538, opened 2020-10-30
Feature/apmlog

chiichang:
AP-2474


raboczi (approved):
Because we don't have branch protection on the ApromoreEE repo and the corresponding commit was pushed there, nobody can build EE until this PR is merged, so I'm going to approve it as-is to unblock everyone else.

We should still retroactively review APMLog for code quality.  There's a lot of commented-out code that should probably be removed, for instance.

PR #544, opened 2020-11-06
Feature/custom calender/ap 2395

nolantellis:
* Added Holiday with tests
* Added logic to calculate duration
* This does not effect any development features.


Apromore-Calendar/src/main/java/org/apromore/calendar/service/CustomCalendarService.java
```diff
@@ -38,68 +40,86 @@
...
+    OffsetTime endTime = OffsetTime.of(LocalTime.of(17, 0), ZoneOffset.UTC);
+
+    CustomCalendar customCalender = createCalendar(description, weekendsOff, startTime, endTime);
+    CalenderModel calenderModel = modelMapper.getMapper().map(customCalender, CalenderModel.class);
+
+    return calenderModel;
+
+  }
+
+  public CalenderModel getCalenderById(Long id) {
```
dreamind: Calendar instead of Calender?

dreamind:
While all the workday and holiday definition is set in UTC, should we add the TimeZone to the CalendarModel?
Is the duration calculated in the UTC context? To account for TimeZone specific quirks (like daylight saving) should we use ZonedDateTime (and perhaps ChronoUnit.between)?

raboczi (changes requested):
As already noted by Ivo, CalenderModel needs to be renamed as CalendarModel.  Additionally, you should configure your IDE to indent with spaces rather than tabs; it seems to keep rewriting swathes of XML.

Apromore-Calendar/src/main/java/org/apromore/calendar/service/CustomCalendarService.java
nolantellis: Corrected.

nolantellis:
> While all the workday and holiday definition is set in UTC, should we add the TimeZone to the CalendarModel?
> Is the duration calculated in the UTC context? To account for TimeZone specific quirks (like daylight saving) should we use ZonedDateTime (and perhaps ChronoUnit.between)?

You are right. I will merge this changes which does not effect anything. I have created another task to convert it to ZonedDateTime

dreamind (approved):

raboczi (approved):
Good enough to merge!

PR #548, opened 2020-11-10
AP-2556 and code cleaning

chiichang:


raboczi (approved):
In Util.java you may want to consider the [java.time.Duration](https://docs.oracle.com/javase/8/docs/api/java/time/Duration.html) class as way to do at least some of the conversion of milliseconds to minutes, hours, and days.  Months and years you'd have to keep doing the current way.

PR #551, opened 2020-11-11
AP-2461 Fix overlapped in tile view

dreamind:


Apromore-Core-Components/Apromore-Portal/src/main/java/org/apromore/portal/dialogController/renderer/SummaryItemRenderer.java
```diff
@@ -268,13 +268,15 @@ protected Listcell renderNA() {
 
     protected Listcell renderOwner(final SummaryType summaryType) {
         Boolean isMakePublic = summaryType.isMakePublic();
-        String label;
+        String owner;
```
nolantellis: Can we make this 
owner = summaryType.getOwner();

Then have only 1 if statement

nolantellis (approved):

PR #553, opened 2020-11-12
AP-2459 Revert to use group id

dreamind:


raboczi (approved):
I can confirm this un-breaks the manager tests, so I'm eager to get it merged to unblock other PRs.

PR #554, opened 2020-11-13
AP-2283 Reskin simulator

dreamind:


raboczi (commented):
Are the modifications to site.properties.template intentional, or is this just a local change that was mistakenly committed?

raboczi (approved):

PR #555, opened 2020-11-13
ZK 9 compatibility

raboczi:
These changes allow the ZK version in the top-level pom.xml to be switched to 9.5.0.1, in which case the server will build and start, but there are styling issues that mean it's not ready for production.  This at least opens the way to debugging it in future.  With the ZK version left at its current value (8.6.0.1) nothing changes.

Other tweaks:
- the portal's web root has to be / or the system won't work; setting site.portal to / sets this directly.  The only reason the current system runs is because the web root is overridden back to / in apromore-core.plan.  Making / the default saves having to do a similar override in Karaf's features.xml.
- the Portal dynamically accesses org.apromore.service.AuthorizationService in its Spring application context, so it needs to explicitly declare that package import
- the Google JSON library is a requirement for Parquet that weirdly isn't explicitly present; for mysterious reasons ZK 8 seems to have previously made this available so things worked by accident.


Apromore-Core-Components/Apromore-Portal/pom.xml
```diff
@@ -68,6 +68,7 @@
                             org.aopalliance.aop,                            
                             org.apromore.security.filter,
                             org.apromore.security.provider,
+                            org.apromore.service,
```
raboczi, line 71: Needed to access AuthorizationService

Apromore-Assembly/Custom-Plugins-Assembly/pom.xml
```diff
@@ -90,6 +90,11 @@
             <version>5.0.3</version>  <!-- matches hadoop-project 3.3.0 -->
             <type>bundle</type>
         </dependency>
+        <dependency>
+            <groupId>com.google.code.gson</groupId>
+            <artifactId>gson</artifactId>
+            <version>2.7</version>
+        </dependency>
```
raboczi, line 97: Needed by Parquet.

raboczi, line 97: Needed by Parquet.

Apromore-Extras/Chiaroscuro-Theme/pom.xml
```diff
@@ -14,7 +14,6 @@
 	<artifactId>ui-theme-compact</artifactId>
 	<version>1.0-SNAPSHOT</version>
 	<properties>
-		<zk.version>8.6.0.1</zk.version>
```
raboczi, line 17: zk.version should be inherited so that it matches the rest of the system.

dreamind (approved):

PR #557, opened 2020-11-15
AP-2455-2476-2506 User Management refinements

dreamind:


raboczi (commented):
I would prefer ComponentUtil not be in Apromore-Commons because it introduces a dependency on ZK.  Other services that depend on Apromore-Commons such as Apromore-Calendar shouldn't require the ZK libraries.

Apromore-Commons/pom.xml
```diff
@@ -66,6 +66,12 @@
 			<version>1.4.2</version>
 			<scope>test</scope>
 		</dependency>
+		<dependency>
+			<groupId>org.zkoss.zk</groupId>
+			<artifactId>zul</artifactId>
+			<version>${zk.version}</version>
+			<optional>true</optional>
+		</dependency>
```
nolantellis: I agree with Simon. This should be in some Generic UI module. For now Portal sounds right for a quick move.

dreamind:
ComponentUtils is now moved under portal

Apromore-Core-Components/Apromore-Portal/src/main/java/org/apromore/portal/common/zk/ComponentUtils.java
```diff
@@ -0,0 +1,69 @@
...
+import org.slf4j.LoggerFactory;
+
+public final class ComponentUtils {
+    private static final String SCLASS_OFF = "ap-state-off";
+    private static final String SCLASS_ON = "ap-state-on";
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(ComponentUtils.class.getName());
+
+    public static void toggleSclass(HtmlBasedComponent comp, boolean newState, String sclassOff, String sclassOn) {
+        String sclass = comp.getSclass();
```
nolantellis, line 36: Not important now, But StringBuilder will be great in this place.

Apromore-Core-Components/Apromore-Portal/src/main/java/org/apromore/portal/common/zk/ComponentUtils.java
```diff
@@ -0,0 +1,69 @@
...
+        }
+        comp.setSclass(sclass);
+    }
+
+    public static void toggleSclass(HtmlBasedComponent comp, String sclassOff, String sclassOn) {
+        String sclass = comp.getSclass();
+        if (sclass == null) {
+            sclass = "";
+        }
+        if (sclass.contains(sclassOn)) {
```
nolantellis, line 55: pass sclass.contains(sclassOn) as argument to the method and remove if else block

nolantellis (approved):

PR #560, opened 2020-11-16
AP-2579

chiichang:


nolantellis (approved):
If possible move the duplicate code in Some Util class in AmpLog Module

PR #562, opened 2020-11-19
AP-2482 Fix about icon in base about plugin

dreamind:


dreamind:
Just an icon fix

raboczi (approved):
Rebuilt and tested on both Core and EE; looks fine.

PR #566, opened 2020-11-25
Sharing feature

franklma:
1. Sharing window in the portal
2. Dashboard list window
3. Filter list window
4. Migration solution for usermetadata table


raboczi (commented):
Can we arrange to meet and have you walk Nolan and me through this?  There's nothing obviously out of place, but since it deal with the common code in the Manager I'd like to understand it a little better first.

Apromore-Clients/portal-model/src/main/java/org/apromore/portal/model/UserMetadataSummaryType.java
```diff
@@ -0,0 +1,99 @@
...
+package org.apromore.portal.model;
+
+public class UserMetadataSummaryType extends SummaryType {
+
+    private String groupName;
+    private String groupUID;
+    private String createdBy;
+    private String updatedTime;
+    private String accessType;
+    private boolean canShare;
```
karthik-balasubramanian, line 31: Please consider renaming to shareable, editable, viewable. Makes the getter/setter readable

Apromore-Core-Components/Apromore-Manager/src/main/java/org/apromore/service/impl/AuthorizationServiceImpl.java
```diff
@@ -44,12 +45,15 @@
     private static final Logger LOGGER = LoggerFactory.getLogger(AuthorizationServiceImpl.class);
 
     private WorkspaceService workspaceService;
+    private UserMetadataService userMetadataService;
 
     @Inject
-    public AuthorizationServiceImpl(final WorkspaceService workspaceService) {
+    public AuthorizationServiceImpl(final WorkspaceService workspaceService, UserMetadataService userMetadataService) {
```
karthik-balasubramanian, line 51: Please make UserMetadataService final as well.

Apromore-Core-Components/Apromore-Portal/src/main/java/org/apromore/portal/dialogController/BaseController.java
```diff
@@ -98,6 +101,14 @@ public SecurityService getSecurityService() {
         return securityService;
     }
 
+    public AuthorizationService getAuthorizationService() {
```
karthik-balasubramanian, line 104: Why isnt the authorizationService injected?

Apromore-Core-Components/Apromore-Portal/src/main/java/org/apromore/portal/dialogController/BaseListboxController.java
```diff
@@ -477,22 +478,41 @@ public void rename() throws InterruptedException {
...
-            arg.put("currentUser", UserSessionManager.getCurrentUser());
-            Window window = (Window) Executions.getCurrent().createComponents("macros/share.zul", null, arg);
-            window.doModal();
+            SummaryType summaryType = (SummaryType) selectedItem;
+            UserType userType = UserSessionManager.getCurrentUser();
+            String userName = userType.getUsername();
+            Group userAsGroup = getSecurityService().getGroupByName(userName);
+            Map<Group, AccessType> groupAccessMap;
+            Integer selectedItemId = summaryType.getId();
+            if (summaryType instanceof ProcessSummaryType) {
```
karthik-balasubramanian, line 498: Please dont use instanceOf.

https://armedia.com/blog/instanceof-avoid-in-code/

Apromore-Database/src/main/java/org/apromore/dao/UsermetadataRepository.java
```diff
@@ -22,9 +22,22 @@
...
 
+import javax.persistence.QueryHint;
+
 @Repository
 public interface UsermetadataRepository extends JpaRepository<Usermetadata, Integer> {
+
+    Usermetadata findById(Integer id);
+//
+//    @Query("SELECT um FROM Usermetadata WHERE um.id = ?1 FOR UPDATE")
+//    @org.springframework.data.jpa.repository.QueryHints(value = {
```
karthik-balasubramanian, line 39: Please remove this commented out query

karthik-balasubramanian:
Please add unit tests for the Controller, Service classes etc.

Apromore-Core-Components/Apromore-Manager/src/main/java/org/apromore/service/impl/AuthorizationServiceImpl.java
```diff
@@ -95,15 +101,30 @@ private AccessType getAccessType(AccessRights accessRights) {
...
+            groupAccessTypeMap.put(g.getGroup(), getAccessType(accessRights));
+        }
+
+        return groupAccessTypeMap;
+    }
+
+    @Override
+    public void saveLogAccessType(Integer logId, String groupRowGuid, AccessType accessType, boolean shareUserMetadata) {
 
         if (!accessType.equals(AccessType.NONE)) {
```
karthik-balasubramanian, line 120: Please change to if(!AccessType.NONE.equals(accessType)) to avoid NPE

Apromore-Core-Components/Apromore-Manager/src/main/java/org/apromore/service/impl/AuthorizationServiceImpl.java
```diff
@@ -120,20 +142,37 @@ public void saveFolderAccessType(Integer folderId, String groupRowGuid, AccessTy
         }
     }
 
+    @Override
+    public void saveUserMetadtarAccessType(Integer userMetadataId, String groupRowGuid, AccessType accessType) {
+
+        if (!accessType.equals(AccessType.NONE)) {
```
karthik-balasubramanian, line 148: same as above

Apromore-Core-Components/Apromore-Manager/src/main/java/org/apromore/service/impl/WorkspaceServiceImpl.java
```diff
@@ -357,8 +357,29 @@ public String saveLogPermissions(Integer logId, String groupRowGuid, boolean has
...
+    @Override
+    public String saveLogAccessRights(Integer logId, String groupRowGuid, boolean hasRead, boolean hasWrite,
+                                      boolean hasOwnership, boolean shareUserMetadata) {
+        Log log = logRepo.findOne(logId);
+        Group group = groupRepo.findByRowGuid(groupRowGuid);
+
+        createGroupLog(group, log, hasRead, hasWrite, hasOwnership);
+
+        Folder parentFolder = log.getFolder();
+        while (parentFolder != null && parentFolder.getId() > 0) {
```
karthik-balasubramanian, line 372: Maybe change this to a recursive call instead of using a while loop

raboczi (approved):
I'm satisfied that this is good enough to merge.

The need to create new classes like Artifact and AccessRights isn't so much a problem of code duplication as indicative that our existing business object model is inadequate and forcing people to work around it.  Until we can find the time to fix the foundations, this sort of workaround is as clean a way as we can manage.

Minor quibble: it'd be nice to standardize on either Usermetadata or UserMetadata.

nolantellis (approved):

Apromore-Core-Components/Apromore-Portal/src/main/java/org/apromore/portal/dialogController/BaseListboxController.java
dreamind: @karthik-balasubramanian I think in this context, instanceof is simply used for type checking to verify allowable actions for the object of interest. Could you suggest an alternative to do that? We can introduce some kind of enum-based typing but instanceof seems to be the most efficient (https://stackoverflow.com/revisions/26514984/10).

karthik-balasubramanian: Using instanceOf operator is not object oriented programming. We should create an interface that has an abstract method, say getAllowableAction that returns the groupAccessMap. The summary type can implement that interface and each summary type will provide the appropriate access type. 

Also we dont need to do premature optimization. One level of indirection is not going to add a lot of performance impact here.

PR #568, opened 2020-11-26
Added an new Timestamp parser

odaghagheleh:


raboczi (changes requested):
The change to site.properties needs to be backed out.

raboczi (approved):
Builds and runs without trouble, imports .parquet as advertised.  Approved!

PR #572, opened 2020-11-29
AP-2593 Update display name for user menu

dreamind:


raboczi (approved):

raboczi (approved):
You could make getDisplayName static since it's purely functional.  I agree with not making it a method on UserType, since UserType is a DTO.

PR #574, opened 2020-11-30
AP-2506 More dirty checking in User management

dreamind:


raboczi (commented):
The code looks fine and it builds and runs.  What do I need to do to test the new behavior?

dreamind:
> The code looks fine and it builds and runs. What do I need to do to test the new behavior?

Switching user and group tabs while there is an unsaved changes should prompt a warning.

raboczi (approved):
Works for me.  Approved!

PR #582, opened 2020-12-01
Ap 2650 Adjustments based on reviews

franklma:


raboczi (approved):
These changes all seem to be pure refactoring that just makes things safer; if the code wasn't broken before, nothing here is going to break it.

PR #587, opened 2020-12-03
AP-2672 Add user duplicate error

dreamind:


raboczi (commented):
We may be able to do better than catching Exception and guessing that maybe a duplicate user was the problem, but it depends on whether the exception SecurityService.createUser throws in this case is informative or not.  If it's something like javax.persistence.EntityExistsException or the Spring Data equivalent, it would make sense to have the more specific catch block which could display more confident advice in the error message.

If the exception SecurityService.createUser is producing doesn't tell us this information, it should probably improved.  Ideally there would be a @throws doc comment on the method SecurityService.createUser documenting what the implementation currently does.

dreamind:
> We may be able to do better than catching Exception and guessing that maybe a duplicate user was the problem, but it depends on whether the exception SecurityService.createUser throws in this case is informative or not. If it's something like javax.persistence.EntityExistsException or the Spring Data equivalent, it would make sense to have the more specific catch block which could display more confident advice in the error message.
> 
> If the exception SecurityService.createUser is producing doesn't tell us this information, it should probably improved. Ideally there would be a @throws doc comment on the method SecurityService.createUser documenting what the implementation currently does.

I've tried to source the exception, but it seems JPA returns a very generic exception.

raboczi:
> I've tried to source the exception, but it seems JPA returns a very generic exception.

Created a Jira issue to revisit this later: https://apromore.atlassian.net/browse/AP-2701

raboczi (approved):

PR #593, opened 2020-12-04
Some minor changes to the logic to support the new log animation

brucenguyen1:
This animation logic module contains the main log-model alignment algorithm. It was created out of research, the code is mostly unchanged after many years. This legacy code/algorithm will be replaced soon, so it is advised not to spend much time on it. 
Some small changes were made in this PR to support the new log animation portal plugin:
- Change the old name space from "de.hpi.bpmn2_0" to "org.apromore.service.loganimation"
- Some classes are copied and renamed: XXX to XXX2. The current log animation will use XXX while the new log animation uses XXX2, e.g. LogAnimationService -> LogAnimationService2, LogAnimationServiceImpl2, AnimationJSONBuilder2. 
- Small changes were made to XXX2 classes while most of the legacy code is retained (not to spend too much time because they will be refactored soon).

This PR will not affect the current log animation, but it must be merged first before other PRs are issued later.


Apromore-Custom-Plugins/Log-Animation-Logic/src/main/java/org/apromore/service/loganimation/LogAnimationService2.java
```diff
@@ -0,0 +1,41 @@
...
+
+   class Log {
+       public String fileName;
+       public XLog   xlog;
+       public String color;
+   }
+
+   public Object[] createAnimation(String bpmn, List<Log> logs) throws Exception;
+   public Object[] createAnimationWithNoGateways(String bpmnWithGateways, String bpmnNoGateways, List<Log> logs) throws Exception;
+}
```
raboczi, line 41: Are there intended to be identical interfaces LogAnimationService and LogAnimationService2, or is one of them a transition holdover?

raboczi (approved):
This builds and runs without breaking the old log animation and seems tidy and well-thought-out, so it's good enough to merge.

I suggest that instead of supporting the old and new implementations by renaming the interface LogAnimationService to LogAnimationService2, do the same technique you previously used when there were Signavio and BPMN.io implementations of the BPMN editor, using OSGi properties and filters to select between the alternatives.  The other XXX2 classes that aren't exposed as interfaces are fine to stay as they are.

If you expect this code to be discarded shortly and it's not worth the effort, I'm happy to leave that to your judgment.  However this seems like a fairly minor change that can be done entirely within context.xml attributes.  It'll save having to change the LogAnimationService2 back once the migration is over.

dreamind (approved):

brucenguyen1:
> This builds and runs without breaking the old log animation and seems tidy and well-thought-out, so it's good enough to merge.
> 
> I suggest that instead of supporting the old and new implementations by renaming the interface LogAnimationService to LogAnimationService2, do the same technique you previously used when there were Signavio and BPMN.io implementations of the BPMN editor, using OSGi properties and filters to select between the alternatives. The other XXX2 classes that aren't exposed as interfaces are fine to stay as they are.
> 
> If you expect this code to be discarded shortly and it's not worth the effort, I'm happy to leave that to your judgment. However this seems like a fairly minor change that can be done entirely within context.xml attributes. It'll save having to change the LogAnimationService2 back once the migration is over.

This is the logic-module which would be used by CE and EE log animation portal plugins. As I wrote in another PR (https://github.com/apromore/ApromoreCore/pull/594), I'm preparing to do the above for the two CE and EE portal plugins similar to Signavio and BPMN.io. This technique is not applied to the logic module as it is unnecessary. It is only temporary during this transition period as we want to keep the old log animation for a while. Once the new log animation is ready, all the old interface (as marked Deprecated) XXX will be removed and the new XXX2 will be come XXX.

PR #594, opened 2020-12-04
Create log-animation-portal-plugin-api module

brucenguyen1:
This api module in Core will be used by log animation module CE and log animation module EE. The CE and EE module will register the same LogAnimationPluginInterface with Spring/OSGi but with a different property key (either "CE" or "EE").

With this API module, other modules using the log animation feature such as PD only need to depend on the API module. When they want specific version either CE or EE, they can query Spring/OSGi with the property key to obtain the right dependency; however, the interface is always LogAnimationPluginInterface which has no impact on their code using either CE or EE version. At runtime, the engine (Spring/Osgi) will return the right bean based on the provided key.

This PR is needed to merge before other PRs to modify the current and create the new log animation module.

### DIFF
--- a/Apromore-Custom-Plugins/Log-Animation-Portal-Plugin-API/pom.xml
+++ b/Apromore-Custom-Plugins/Log-Animation-Portal-Plugin-API/pom.xml
@@ -1,0 +1,56 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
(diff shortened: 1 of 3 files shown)


raboczi (commented):
I am enthusiastic about separate API bundles for services, such that log-animation-portal-plugin depends just on this API and not on the whole of log-animation-logic.

At the moment it doesn't look like it needs merging since it doesn't interact with anything outside its own directory: i.e. it's not a module for building from the parent pom.xml, and the other log-animation components evidently don't refer to it yet.  It's certainly not going to break anything if you merge it now, but it seems like it could wait until it's actually being used.  Or at least, built.

raboczi (approved):

dreamind (approved):

PR #599, opened 2020-12-07
AP-2062: Remove PortalPluginResourceServlet

raboczi:
This PR removes PortalPluginResourceServlet.  It must only be merged after its companion PR https://github.com/apromore/ApromoreEE/pull/130 in ApromoreEE.

PortalPluginResourceServlet was used to access static web content stored in a PortalPlugin bundle, and specifically for the SVG icons for the portal plugins appearing on the main menu. If this PR is broken, the icons on the main menubar will have broken links or appear as the wrong image.

The old scheme for the icons has been replaced with BundleWebContentServlet, which is an improved version of SimpleWebContentServlet. (I plan to remove SimpleWebContentServlet in a future PR.) Doing this requires that all the icons have unique classpaths, so I've moved various identically-named icon.svg files into per-component subdirectories to distinguish them, and added getIconPath methods.


Apromore-Plugins/plugin-core/portal/api/src/main/java/org/apromore/plugin/portal/BundleWebContentService.java
```diff
@@ -0,0 +1,145 @@
...
+    public InputStream getResourceAsStream(String path) {
+        assert hasResource(path);
+        try {
+            return bundle.getResource(path).openStream();
+
+        } catch (IOException e) {
+            throw new Error("Unable to read " + path, e);
+        }
+    }
+}
```
brucenguyen1, line 145: Reviewing the portal-plugin-api as a whole where this class is located, this getResourceAsStream() method duplicates the getResourceAsStream() method of the PortalPlugin interface. The plugin resource API in this module needs refactoring before it gets more difficult to do: e.g. the API exposed via PortalPlugin, DefaultPortalPlugin and the API exposed via these WebContentService/BundleWebContentService/SimpleWebContentService.

brucenguyen1: Looking at other classes in this module (portal-plugin-api), they need a cleanup because the classes added over time are not cohesive (i.e. serving the same API). 
For example:
- These classes should not be here: MainControllerInterface, SessionTab, PortalSelection. They should be part of the Portal.
- Enum Level is unused.
- FileImporterPlugin: how is it related to the PortalPlugin API? 
- PortalPlugin, DefaultPortalPlugin, WebContentServcie, BundleWebContentService, SimpleWebContentService: the first two classes expose the Plugin API including the plugin resource loading API (so they fit in this module), while the last 3 classes are for resource loading API. However, these two groups are not related to each other, i.e. any extension, association, use relationship?

Again, while modifying and adding more classes in this module, I suggest to do a bit refactoring to clean it before it gets more difficult to do: move classes to the right place, renaming methods, move methods to avoid duplication, etc.

brucenguyen1 (changes requested):

Apromore-Custom-Plugins/Log-Animation-Portal-Plugin/src/main/java/org/apromore/plugin/portal/loganimation/LogAnimationPlugin.java
```diff
@@ -79,6 +79,11 @@ public void setGroupLabel(String groupLabel) {
         this.groupLabel = groupLabel;
     }
 
+    @Override
+    public String getIconPath() {
+        return "loganimation/icon.svg";
+    }
+
```
brucenguyen1, line 86: The same comment for the same PR in EE repo: this behavior should be made default, only plugin with exception in icon or icon path should override this behavior. This would reduce much boilerplate code and missing errors.

Apromore-Custom-Plugins/Merge-Portal-Plugin/src/main/resources/META-INF/spring/osgi-context.xml
```diff
@@ -41,4 +41,11 @@
     <osgi:reference id="mergeService" interface="org.apromore.plugin.merge.logic.MergeService"/>
     <osgi:reference id="domainService" interface="org.apromore.service.DomainService"/>
 
+    <osgi:service interface="org.apromore.plugin.portal.WebContentService">
+        <beans:bean class="org.apromore.plugin.portal.BundleWebContentService">
+            <beans:constructor-arg ref="bundleContext"/>
+            <beans:property name="include" value="/merge/icon.svg"/>
+        </beans:bean>
+    </osgi:service>
+
```
brucenguyen1, line 50: The icon resource location "/merge/icon.svg": can it be configured in one place instead of leaving two places: here and MergePlugin.getIconPath()?

Apromore-Plugins/plugin-core/portal/api/src/main/java/org/apromore/plugin/portal/BundleWebContentService.java
```diff
@@ -0,0 +1,145 @@
...
+ * &lt;osgi:service interface="org.apromore.plugin.portal.WebContentService"&gt;
+ *   &lt;beans:bean class="org.apromore.plugin.portal.BundleWebContentService"&gt;
+ *     &lt;beans:constructor-arg ref="bundleContext"/&gt;
+ *     &lt;beans:property name="include" value="/fancyviewer/.*\.(gif|jpg|png|svg)"/&gt;
+ *     &lt;beans:property name="exclude" value="WEB-INF/.*|.*\.class|.*&#47;"/&gt;
+ *   &lt;/beans:bean&gt;
+ * &lt;/osgi:service&gt;
+ * </pre>
+ */
+public class BundleWebContentService implements WebContentService {
```
brucenguyen1, line 62: I can see this BundleWebContentService was created to make it possible to declare this bean in the context.xml file, as explained in the class header comment above. However, it is not simple to write the EL expression for include/exclude considering a (complex) plugin could have various resources and hierarchy of resource organization, and there is much duplicate boilerplate code across plugins with little difference. Instead, I suggest defining a standard resource structure for plugins (e.g. a "resources" folder containing all resource files in a plugin) and write a default implementation class for loading resources in a standard way, and so any plugins would inherit this behavior instead of implementing duplicated boilerplate code. That's what 80-90% plugins would need, those plugins need exception way of loading resource can override the default behaviors.

Apromore-Custom-Plugins/Merge-Portal-Plugin/src/main/java/org/apromore/plugin/merge/portal/MergePlugin.java
```diff
@@ -111,6 +111,11 @@ public String getGroupLabel(Locale locale) {
         return "Redesign";
     }
 
+    @Override
+    public String getIconPath() {
+        return "merge/icon.svg";
+    }
+
```
brucenguyen1, line 118: What happens if two different plugins have the same resource string, e.g. "merge/icon.svg", but the resources are actually different, e.g. they are different icons. Is it the "merge" here indicates that it must be unique key only used for the Merge-Portal plugin? If so, is there anyway in the current implementation to detect/avoid this resource location clash?

Apromore-Custom-Plugins/File-Portal-Plugin/src/main/resources/META-INF/spring/context.xml
```diff
@@ -75,4 +75,11 @@
         <beans:bean class="org.apromore.plugin.portal.file.DeleteSelectionPlugin"/>
     </osgi:service>
 
+    <osgi:service interface="org.apromore.plugin.portal.WebContentService">
+        <beans:bean class="org.apromore.plugin.portal.BundleWebContentService">
+            <beans:constructor-arg ref="bundleContext"/>
+            <beans:property name="include" value=".*\.svg"/>
+        </beans:bean>
+    </osgi:service>
+
```
brucenguyen1, line 84: Why the resource location pattern here doesn't need any key identifying the plugin as shown in other plugins, like "loganimation/icon.svg", "merge/icon.svg", "/similaritysearch/icon.svg"

Apromore-Plugins/plugin-core/portal/api/src/main/java/org/apromore/plugin/portal/BundleWebContentService.java
raboczi: Refactoring the Portal APIs is a worthwhile endeavor, but it's somewhat beyond the scope I had in mind for this particular PR.  For now I just want BundleWebContentService to replace PortalPluginResourceServlet and all the WebContentService implementations (SimpleWebContentService, BPMNEditorWebContentService, PDWebContentService...)

Apromore-Custom-Plugins/Merge-Portal-Plugin/src/main/resources/META-INF/spring/osgi-context.xml
raboczi: This is a (somewhat) unusual case where the ONLY web content for the merge plugin is its icon.  In the more general case, the icon path and the include regexp wouldn't match like this.

Apromore-Custom-Plugins/Merge-Portal-Plugin/src/main/java/org/apromore/plugin/merge/portal/MergePlugin.java
raboczi: Nope.  If there's contention, the first WebContentService checked is the one that wins.  Detection wouldn't be terribly hard, but there would be a slight performance hit to keep checking after the first match.

Apromore-Custom-Plugins/File-Portal-Plugin/src/main/resources/META-INF/spring/context.xml
raboczi: The icons in the file portal plugin all have unique names (e.g. model-edit.svg, upload-file.svg) rather than ever being called just icon.svg.  It's a bit of an artifact of them being originally part of the Portal rather than starting their lives as a plugin.  It wouldn't be any great effort to copy them into a file/ subdirectory to use the same convention as the others to avoid contention.

PR #600, opened 2020-12-07
AP-2583: Update PD to run two versions of log animation portal plugins.

brucenguyen1:
This PR updates PD to run two log animation portal plugins. There are two play buttons on the toolbar of PD: first one is the current log animation (Core repo), new one is the log animation EE version (EE repo).

Detail changes:
- Configuration (pom, context): add dependency on the log animation API module, add to Spring context.xml to query two beans with 'CE' and 'EE' key values.
- Changes to Controller, Plugin classes to use these two CE and EE beans: add AnimationController2 for connecting to the EE bean. It is mostly the same as AnimationController using the CE bean. When the EE version is complete, the AnimationController2 will replace AnimationController and there will be only one play button left.
- ZUL file: add one more play button.

### DIFF
--- a/Apromore-Custom-Plugins/Process-Discoverer-Portal-Plugin/src/main/java/org/apromore/plugin/portal/processdiscoverer/controllers/AnimationController.java
+++ b/Apromore-Custom-Plugins/Process-Discoverer-Portal-Plugin/src/main/java/org/apromore/plugin/portal/processdiscoverer/controllers/AnimationController.java
@@ -29,6 +29,7 @@ import org.apromore.processmining.models.graphbased.directed.bpmn.BPMNDiagram;
 import org.apromore.processmining.models.jgraph.ProMJGraph;
 import org.apromore.processmining.plugins.bpmn.BpmnDefinitions;
 import org.zkoss.zk.ui.event.Event;
+import org.zkoss.zul.Messagebox;
 
 public class AnimationController extends AbstractController {
     public AnimationController(PDController controller) {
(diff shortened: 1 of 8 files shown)


raboczi (approved):

Apromore-Custom-Plugins/Process-Discoverer-Portal-Plugin/src/main/resources/processdiscoverer/zul/processDiscoverer.zul
```diff
@@ -205,6 +205,7 @@
         <n:span class="ap-action-sep"></n:span>
         <!-- <button ca:data-t="animate" id="animate" w:onClick="Ap.pd.animate()" tooltiptext="Animate" sclass="ap-icon ap-icon-play-circle"/> -->
         <button ca:data-t="animate" id="animate" tooltiptext="Animate" sclass="ap-icon ap-icon-play-circle"/>
+        <button ca:data-t="animate" id="animate_new" tooltiptext="Animate NEW" sclass="ap-icon ap-icon-play-circle"/>
```
nolantellis, line 208: is it possible to put like some red '*' indicating its a new log annimation. Just a simple quick differ between the 2 log animation. Nothing jazzy.

Apromore-Custom-Plugins/Process-Discoverer-Portal-Plugin/src/main/java/org/apromore/plugin/portal/processdiscoverer/controllers/AnimationController.java
```diff
@@ -1,90 +1,96 @@
...
+            Messagebox.show("Log Animation Plugin CE version is not available.");
+            return;
+        }
+        
+        Abstraction abs = parent.getOutputData().getAbstraction();
+        if (abs.getLayout() == null) {
+            throw new MissingLayoutException("Missing layout of the process map for animating");
+        }
+        
+        if (parent.getBPMNMode()) {
```
nolantellis, line 55: it would be gr8 to move it to just 1 execute , with checks based on parameters, I see the execute method is almost the same.
But this is good if the intention is to delete the old execute method eventually

nolantellis (approved):

Apromore-Custom-Plugins/Process-Discoverer-Portal-Plugin/src/main/java/org/apromore/plugin/portal/processdiscoverer/controllers/AnimationController.java
brucenguyen1: Ok, will make sure to remove duplication when making changes to the other execute method (not yet updated).

PR #601, opened 2020-12-07
Add support to rename only if authorized

nolantellis:


raboczi (approved):
This looks fine.  One quibble: did you intentionally reindent BaseListboxController?  Eventually I'd like to enable checkstyle and ban tab characters.

franklma (changes requested):
A user with VIEWER access type can't rename log, process and folder as expected. However, there is casualty that user with proper access type can not rename a folder.

nolantellis:
> A user with VIEWER access type can't rename log, process and folder as expected. However, there is casualty that user with proper access type can not rename a folder.

This is in dev too, Have fixed it

franklma (approved):

PR #604, opened 2020-12-08
Added ristriction to copy/cut/paste

nolantellis:
* User can copy paste From Readonly Folder
* USer cannot Cut from Readonly Folder.
* User cannot paste in Readonly Folder.


franklma (approved):
works as expected.

PR #612, opened 2020-12-11
AP-2722 Update ui.properties to allow percentage

dreamind:


raboczi (approved):
You might just want to double-check that the validation message is right; I think it previously said "...space and ." and you presumably intend it to say something like "space, percent, and period".

PR #616, opened 2020-12-15
Feature/ap 2300

odaghagheleh:
This PR has a counterpart in ApromoreEE https://github.com/apromore/ApromoreEE/pull/140 which depends on the new APIs introduced here.


raboczi (commented):
I've piggybacked on this branch to add the improved log upload to the REST endpoint by using the new APIs Own created.  It still works as it used to for XES upload, but the new behavior for CSV and parquet uploads isn't yet working.  I'm pretty sure the wait for the 7.19 code freeze will take longer than the fix.

raboczi:
It's now possible to POST both CSV and parquet files, so this merge isn't blocked on functional grounds.

Just to preempt other reviewers, the following improvements should happen, either before this PR merges or as subsequent issues:
- the temporary files used for parquet uploads shouldn't linger
- POSTed logs are always owned by the uploading account; there should be a way to override this
- error reporting for unsuccessful POST requests is uninformative; it should at least tell you which log attributes didn't have HTTP headers specifying their type

raboczi (approved):
This code has be tested both my myself and Karthik and the people working on landing the 7.19 branch have requested its bug fixes, so it's approved to merge earlier than expected.

PR #617, opened 2020-12-15
AP-2832

chiichang:
- Incorrect arc dur val caused by mis-assigned overlapped events to activities.
- Fixed with new unit test added

### DIFF
--- a/Apromore-Extras/APMLogModule/src/main/java/org/apromore/apmlog/filter/PTrace.java
+++ b/Apromore-Extras/APMLogModule/src/main/java/org/apromore/apmlog/filter/PTrace.java
@@ -398,6 +398,8 @@ public class PTrace implements Comparable<PTrace>, ATrace {
         if (fromIndex == eventList.size() - 1) proceed = false;
         if (baseLife.equals("complete")) proceed = false;
 
+        boolean foundStart = baseLife.equals("start");
+
         if (proceed) {
             for (int i = fromIndex + 1; i < eventList.size(); i++) {
                 if (!markedIndexes.contains(i)) {
@@ -415,7 +417,14 @@ public class PTrace implements Comparable<PTrace>, ATrace {
 
                             break;
                         } else {
-                            eventIndexList.add(i);
+                            if (lifecycle.equals("start")){
+                                if (!foundStart) {
+                                    eventIndexList.add(i);
+                                    foundStart = true;
+                                }
+                            } else {
+                                eventIndexList.add(i);
+                            }
                         }
                     }
                 }
--- a/Apromore-Extras/APMLogModule/src/test/java/org/apromore/apmlog/AttributeArcDurationTest.java
+++ b/Apromore-Extras/APMLogModule/src/test/java/org/apromore/apmlog/AttributeArcDurationTest.java
@@ -36,6 +36,8 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
 
+import static org.junit.Assert.assertEquals;
+
 public class AttributeArcDurationTest {
     public static void testRetain1(APMLog apmLog, APMLogUnitTest parent) throws UnsupportedEncodingException {
         FilterType filterType = FilterType.ATTRIBUTE_ARC_DURATION;
@@ -102,11 +104,10 @@ public class AttributeArcDurationTest {
             System.out.println(trace.getCaseId());
         }
 
-        if (hasC1 || hasC2 || hasC3 || !hasC4) {
-            throw new AssertionError("TEST FAILED. RESULT TRACE LIST MISMATCH.\n");
-        } else {
-            parent.printString("'Attribute arc duration (1)' test PASS.\n");
-        }
+        assert !hasC1;
+        assert !hasC2;
+        assert !hasC3;
+        assert hasC4;
     }
 
     public static void testRetain2(APMLog apmLog, APMLogUnitTest parent) throws UnsupportedEncodingException {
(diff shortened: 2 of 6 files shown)


Apromore-Extras/APMLogModule/src/test/java/org/apromore/apmlog/AttributeArcDurationTest.java
```diff
@@ -252,4 +255,39 @@ public static void testRetain3(APMLog apmLog, APMLogUnitTest parent) throws Unsu
...
+    public static double getTotal(DoubleArrayList doubleList) {
+        if (doubleList.size() > 0) {
+            double sum = 0;
+            for (int i = 0; i < doubleList.size(); i++) {
+                sum += doubleList.get(i);
+            }
+            return sum;
+        }
+        return 0;
+    }
```
raboczi: I think you may find that DoubleArrayList has a .sum() method which makes this method unnecessary.

Apromore-Extras/APMLogModule/src/test/java/org/apromore/apmlog/AttributeArcDurationTest.java
```diff
@@ -252,4 +255,39 @@ public static void testRetain3(APMLog apmLog, APMLogUnitTest parent) throws Unsu
...
+        }
+    }
+
+    public static double getAverage(DoubleArrayList doubleList) {
+        if (doubleList.size() > 0) {
+            double sum = getTotal(doubleList);
+            return doubleList.size() > 1 ? sum / doubleList.size() : sum;
+        }
+        return 0;
+    }
```
raboczi: I think there's a provided DoubleArrayList.average method to do this without an extra method.

Apromore-Extras/APMLogModule/src/test/java/org/apromore/apmlog/AttributeArcDurationTest.java
```diff
@@ -252,4 +255,39 @@ public static void testRetain3(APMLog apmLog, APMLogUnitTest parent) throws Unsu
...
+        DurSubGraph subGraph = pLog.getAttributeGraph().getNextValueDurations("concept:name", "A");
+        DoubleArrayList durList = subGraph.getValDurListMap().get("A");
+        double avg = getAverage(durList);
+        double twoHours = 1000 * 60 * 60 * 2d;
+        if (avg != twoHours) {
+            throw new AssertionError("TEST FAILED. RESULT MISMATCH.\n expected value=" + twoHours +
+                    "; output value=" + avg);
+        } else {
+            parent.printString("'Attribute arc duration (4) - Average duration' test PASS.\n");
+        }
```
raboczi: JUnit has various assertion methods which could simplify test code like this, something like:

```
import static org.junit.Assert.assertEquals;
```
```
assertEquals(1000 * 60 * 60 * 2d, getAverage(durList));
```

raboczi (commented):
Can we arrange a quick walkthrough of this code so that I'm clear about the problem it addresses?  I'm guessing it's about the duration of overlapping events but it doesn't actually mention in the test comments, etc.  This is pretty short so it ought to be quick.

brucenguyen1:
> Can we arrange a quick walkthrough of this code so that I'm clear about the problem it addresses? I'm guessing it's about the duration of overlapping events but it doesn't actually mention in the test comments, etc. This is pretty short so it ought to be quick.

Yes, I tried to understand the changes to do a review but had no clues. The changes in this PR are only a few but the existing code is much bigger, a quick walkthrough to share how things are organized and related here would be helpful to understand why these changes were made to fix the issue, and so why the unit tests were written to cover this scenario.

Apromore-Extras/APMLogModule/src/test/java/org/apromore/apmlog/AttributeArcDurationTest.java
chiichang: I'm not sure where you want to place that code. Please edit it directly.

Apromore-Extras/APMLogModule/src/test/java/org/apromore/apmlog/AttributeArcDurationTest.java
chiichang: updated

Apromore-Extras/APMLogModule/src/test/java/org/apromore/apmlog/AttributeArcDurationTest.java
chiichang: removed

chiichang:
> > Can we arrange a quick walkthrough of this code so that I'm clear about the problem it addresses? I'm guessing it's about the duration of overlapping events but it doesn't actually mention in the test comments, etc. This is pretty short so it ought to be quick.
> 
> Yes, I tried to understand the changes to do a review but had no clues. The changes in this PR are only a few but the existing code is much bigger, a quick walkthrough to share how things are organized and related here would be helpful to understand why these changes were made to fix the issue, and so why the unit tests were written to cover this scenario.

The algorithm starts counting the events not from "start" but from any lifecycle:transition value, which means in the old code, to get the events of an activity, it took all the events that have same concept:name  (including "start") after the first event until it reaches a "complete" or the end of the case. 

That causes problem and is explicitly shown in the arc duration filter (the values derived from DurSubGraph). Hence, the unit test is just to test the value from the DurSubGraph.

Apromore-Extras/APMLogModule/src/test/java/org/apromore/apmlog/AttributeArcDurationTest.java
raboczi: Done!

raboczi (approved):
I'm satisfied that my suggestions have been addressed, and that this builds, passes its tests, and runs.

Apromore-Extras/APMLogModule/src/main/java/org/apromore/apmlog/filter/PTrace.java
```diff
@@ -415,7 +417,14 @@ private AActivity getActivity(int index, int fromIndex, IntArrayList markedIndex
...
+                                if (!foundStart) {
+                                    eventIndexList.add(i);
+                                    foundStart = true;
+                                }
+                            } else {
+                                eventIndexList.add(i);
+                            }
                         }
                     }
                 }
```
brucenguyen1, line 430: This code is duplicated with the code above in LogFactory.java (getActivity). Can we do a bit refactoring here to remove this duplication?

PR #619, opened 2020-12-15
AP-2506 Incorrect dirty flag when arranging group member

dreamind:


Apromore-Custom-Plugins/User-Admin-Portal-Plugin/src/main/java/org/apromore/plugin/portal/useradmin/UserAdminController.java
```diff
@@ -1144,7 +1144,7 @@ public void onClickRetractUser() {
                 nonAssignedUserList.reset();
                 assignedUserList.reset();
             }
-            isGroupDetailDirty = false;
+            isGroupDetailDirty = true;
```
brucenguyen1, line 1147: Not in the scope of this PR, but I saw the access modifier for isGroupDetailDirty (and many other instance variables) are by default (protected), they are visible to all classes in the same package. Can we set them to private as Java supports this aspect very well (the most restrictive information hiding principle)? It will help avoid potential troubles.

Apromore-Custom-Plugins/User-Admin-Portal-Plugin/src/main/java/org/apromore/plugin/portal/useradmin/UserAdminController.java
```diff
@@ -1165,7 +1165,7 @@ public void onClickAssignUser() {
...
```
brucenguyen1, line 1168: The naming "isGroupDetailDirty" is quite impressive, but a more meaningful name would be "isGroupDetailChanged" (and so it needs to be saved if it's true), as I can understand it easily. I would choose the most simple and meaningful names rather than impressiveness.

brucenguyen1 (approved):
Approved with comments.

Apromore-Custom-Plugins/User-Admin-Portal-Plugin/src/main/java/org/apromore/plugin/portal/useradmin/UserAdminController.java
dreamind: I think "dirty" flag is common term to indicate unsaved changes in many JS libraries.
https://github.com/ngneat/dirty-check-forms
https://www.npmjs.com/package/jquery.dirty

Apromore-Custom-Plugins/User-Admin-Portal-Plugin/src/main/java/org/apromore/plugin/portal/useradmin/UserAdminController.java
dreamind: Will fix it

PR #620, opened 2020-12-16
AP-2532 Fix bpmnio issue with native prop in Safari

dreamind:


Apromore-Core-Components/Apromore-BPMNEditor/src/main/resources/bpmneditor/editor/bpmnio/bpmn-modeler.development.js
```diff
@@ -276,7 +276,7 @@ function forEach(collection, iterator) {
   var convertKey = isArray(collection) ? toNum : identity;
 
   for (var key in collection) {
-    if (has(collection, key)) {
+    if (has(collection, key) && key !== 'remove') {
       val = collection[key];
```
brucenguyen1, line 280: Hi Ivo, it could be very tricky to find out where to fix this problem, but is there any other way rather than modifying directly into the bundle file of bpmn.io? For example, the bpmn.io team do support and answer question, did we search in their forum or their help? In case we have to customize bpmn.io itself, can we modify their source rather than the final bundle file bpmn-modeler.development.js (this bundle file was built from source files)? This is because we want to keep it updated with their release, and keep it manageable what are our customization if need to.

dreamind: Yes, this is a temporary solution before we push for the proper one.

brucenguyen1: Please consider whether this is a major and critical issue (showing the editor on Safari) to do this. It would be ok if it's an urgent and serious error from a customer perspective. If not, then my recommendation is taking some time to do the proper solution as you said.

dreamind: The issue is Safari user can't save BPMN model. The bundle is generated from https://github.com/apromore/QBPSimulator/.

raboczi: The first number I could find online suggested that Safari's market share is ~10% for desktop users.  So that means 10% of our users will be mysteriously unable to work with BPMN.  That seems an unacceptable regression in a release version of the software.

raboczi (approved):
Fixes the problem in 20 characters.  Approved.

raboczi:
Created a Jira issue https://apromore.atlassian.net/browse/AP-2869 to revisit this so that we have a better solution than manually patching every new build from Javascript of BPMN.io.

PR #621, opened 2020-12-17
Remove old file formats from native_type table

franklma:


raboczi (approved):
Changing the contents of this table is more like configuration than a code change.  It achieve the goal of removing the unsupported process model types from the upload dialog, builds without issue and runs correctly.  I approve this merge.

I had a quick search through the code.  If we really wanted to root all all references to AML/EPML/PNML/XPDL the one place I spotted some references to the old alternative process model formats is in org.apromore.portal.dialogController.BaseController.  There's also test code (org.apromore.service.impl.ProcessServiceImplUnitTest) which doesn't depend on the database and so isn't broken by this change.  It seems advisable to leave it in there for the moment; someone might theoretically want to resurrect some ApromoreCE someday and configure the native_type table accordingly.  Just before a code freeze is probably not the idea time to do this sort of cleanup.

franklma:
Created a following up card in 7.20:
https://apromore.atlassian.net/browse/AP-2862

PR #622, opened 2020-12-17
AP-2861 Check duplicate group name

dreamind:


raboczi (approved):
Ideally createGroup would be throwing a documented exception specifying that a duplicate was the read it failed, but given the SecurityService as it is, this seems reasonable.

PR #623, opened 2020-12-17
Ap 2780 New share logic

franklma:
Implementation of new share logic.

1. OWNERSHIP OF LOGS AND ARTIFACTS: An owner of a log owns ALL artifacts associated with a log. No artifact associated with a log is hidden from the eyes of its owner. The owner of a log can do whatever they want with the artifacts of a log, 100%, no exceptions!

2. ADD ARTIFACT: An owner or an editor of a log can add new artifacts to the log. A viewer of a log cannot add artifacts to it.

3. DELETE/RENAME/EDIT ARTIFACT: An owner or an editor of a log can delete, rename or edit any artifact of the log. A viewer of a log cannot delete, rename nor add artifacts to a log.

4. SHARE LOG: An owner of a log can share the log. An editor or a viewer of a log cannot share the log. There are two types of sharing in “view” mode (Only Full view sharing is implemented in 7.19):

- > Full view sharing - a full viewer can see all artifacts associated with the log at all times (incl. new artifacts that may be added later) while they retain their full view right on the log. 
- > Restricted view sharing - a restricted viewer of a log can only see and use artifacts to which he/she has been granted explicit access by an owner.

5. ACCESS RIGHTS FOR MULTI-LOG ARTIFACTS. When an artifact A is multi-log (e.g. a multi-log dashboard template), the access right of a user U on artifact A is the most restrictive of the access rights of user U on artifact A across all the logs associated to A. For example, if an artifact A is associated with the pair of logs {L1, L2}, and a user U has owner access to L1 and edit access to L2, then user U cannot share artifact A but they can edit artifact A. This means that to be able to create a multi-log artifact, a user must be owner or editor of each of the logs covered by the artifact.

6. Changes in Portal's share dialogue, use new logic when identifying current user's access rights to a Log/Process/Folder.

Document in confluence:
https://apromore.atlassian.net/wiki/spaces/AP/pages/594804754/New+Share+an+improvement+over+the+current+Share


Apromore-Database/src/test/java/org/apromore/dao/jpa/usermetadata/UserMetadataServiceUnitTest.java
```diff
@@ -0,0 +1,172 @@
...
+        assertThat(savedTestUser.getType()).isEqualTo(testUser.getType());
+    }
+
+    @Test
+    public void testSaveGroup() {
+        // given
+        Group userGroup = builder.withGroup("testGroup1", "GROUP").buildGroup();
+
+        // when
+        Group savedUSerGroup = groupRepository.saveAndFlush(userGroup);
```
dreamind: savedUserGroup instead of savedUSerGroup

Apromore-Core-Components/Apromore-Manager/src/test/java/org/apromore/builder/UserManagementBuilder.java
```diff
@@ -87,15 +104,15 @@ public UserManagementBuilder withUserMetaDataType(String type, int version, bool
 	        return this;
 
 	}
-	
+
```
dreamind, line 107: We probably need to choose between Usermetadata or UserMetadata or UserMetaData.

dreamind (approved):

Apromore-Core-Components/Apromore-Manager/src/main/java/org/apromore/service/helper/UIHelper.java
```diff
@@ -252,15 +249,13 @@ public SummariesType buildLogSummaryList(String userId, Integer folderId, Intege
     }
 
     @Override
-    public UserMetadataSummaryType buildUserMetadataSummary(Group group, Usermetadata usermetadata,
-                                                            AccessType accessType) throws UserNotFoundException {
+    public UserMetadataSummaryType buildUserMetadataSummary(String username, Usermetadata usermetadata,
+                                                            AccessType accessType) {
```
raboczi, line 253: Since buildUserMetadataSummary no longer throws UserNotFoundException, what is the new behavior when an invalid username is passed?

raboczi (approved):
Given others vouch for its functional behavior, I have no objections at the code level to merging this.

PR #626, opened 2020-12-19
AP-2870: JDK 11+ support

raboczi:
This PR makes it possible for Apromore on JDK 11 to build (albeit with tests skipped) and run (albeit only on Karaf, not Virgo).

The project still treats the source code as Java 8 and targets a Java 8 JVM, so theoretically the bundles are unchanged.  The difference is that a JDK 11 environment won't contain the various javax.* packages which were dropped from Java SE after version 11.  By adding a profile to the top-level pom.xml which adds this packages as explicit additional dependencies when Maven detects a JDK 11+ environment, the build will continue to work.

While this change makes building and running under JDK 11 possible, it only does so with tests suppressed; there is an issue in the test suite for Apromore-Calendar under JDK 11 related to the modelmapper library.  I'd support merging even with this bug present, since it'll make it much easier for someone else to subsequently work on fixing it.  Since the profile is inactive under our normal JDK 8 environment, it does no harm beyond cluttering the pom.xml.

A secondary change is that the pom.xml for Test-Tools inherits from the top-level pom.xml.  This was done to inherit settings for the compiler.  Previously the compiler would default to assuming Java 1.5 source code; later JDKs can only go as far back as Java 1.6 so this would fail to build.  A spammy side effect is that the newly-inherited license plugin settings have reformatted all the source code in Test-Tools.  No manual edits have been made to those files.

The ApromoreEE repository inherits these changes from ApromoreCore and should build on JDK 11 without itself requiring any change.


raboczi:
I realize this PR looks pretty sketchy -- requiring tests to be disabled!  However it's a prerequisite to actually fixing builds under JDK 11.  Until this is in, it's hard to assign people to troubleshoot the things that don't work.

The profile is inactive under our normal JDK 8 environment, so the only hard it does in the meantime is cluttering the pom.xml.

raboczi:
If the Karaf PR https://github.com/apromore/ApromoreCore/pull/632 lands before this one does, the following code will need to be added to `core-assemblies/pom.xml`:
```
<profiles>
  <!-- JDK 11 removed many libraries from the core, so they become dependencies -->
  <profile>
    <id>jdk11-extra-libs</id>
    <activation>
      <jdk>[11,)</jdk>
    </activation>
    <dependencies>
      <dependency>
        <groupId>com.sun.activation</groupId>
        <artifactId>javax.activation</artifactId>
        <version>1.2.0</version>
      </dependency>
    </dependencies>
  </profile>
</profiles>
```

raboczi:
> If the Karaf PR #632 lands before this one does, the following code will need to be added to `core-assemblies/pom.xml`:

PR #632 has merged and this modification have been added.

raboczi:
This PR has bitrotted to the point where it'll be easier to redo it than to perform the merge to salvage it.  Closing.

PR #627, opened 2020-12-20
Move the third-party SVG library (Batik) from EE to Core

raboczi:
This PR is a requirement for its counterpart PR https://github.com/apromore/ApromoreEE/pull/148 in ApromoreEE.

This moves the bundled third-party image/document format converters in docx4j-osgi from EE to Core.
It allows the removal of the older version of Batik previously embedded within the BPMN editor's PDF plugin.

Additionally, the PDF plugin is rewritten to avoid the unreliable ServletContext.getRealPath method.  Rather than modifying the WAR directory (which isn't guaranteed to exist) to serve the PDFs for "free", the converted PDFs are written into the temporary file area and accessed by a new TemporaryFileServlet.

The changes to the PDF plugin involved improving the ResourceServlet's support for pluggable servlets; their servlet patterns now support limited wildcarding with a leading and/or trailing "*".

### DIFF
--- a/Apromore-Assembly/Custom-Plugins-Assembly/pom.xml
+++ b/Apromore-Assembly/Custom-Plugins-Assembly/pom.xml
@@ -83,6 +83,92 @@
         </dependency>
         -->
 
+        <!-- Required by the BPMN editor's PDF export plugin -->
+        <dependency>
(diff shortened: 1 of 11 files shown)


chiichang (approved):
Built and run without any problem. Library works correctly.

PR #631, opened 2020-12-23
Bumped version.number in site.properties to 7.20-development

raboczi:
Manually bump the version.number in site.properties to 7.20-development so that the About box is accurate.


chiichang (approved):
built and run without problem. Version number has shown correctly.

PR #632, opened 2020-12-23
AP-2522: Karaf assembly of Apromore Core

raboczi:
This PR is a prerequisite for https://github.com/apromore/ApromoreEE/pull/151 in ApromoreEE, but it's safe to merge on its own or to use with the development branch of ApromoreEE.

This PR adds an alternative assembly of Apromore using Apache Karaf in parallel to the existing one on Eclipse Virgo.
- Build the source as normal using `mvn clean install` from the top level.
  - If you pass the flag `-Dkaraf.archive.zip=true` then the Karaf server will be compressed and installed in your local Maven repo as `mvn:org.apromore/apomore-core/7.20-SNAPSHOT/zip`.  I've disabled this for regular development because it slows the regular build-test-debug cycle.
  - `-Dkaraf.archive.tar.gz=true` will similarly create and install a gzip'ed TAR archive.
- Startup varies slightly by OS:
  - On unix-like systems, execute `core-assemblies/apromore-core/target/assembly/bin/karaf`
  - On macOS the default Java Swing look-and-feel is a proprietary Apple class which isn't available in the stock OSGi framework, causing the ProM libraries to fail.  The simplest workaround is to add the following to your environment:
    ```
    export JAVA_OPTS=-Dswing.defaultlaf=javax.swing.plaf.metal.MetalLookAndFeel
    ```
    Then use the same script as other unixen.
  - On Windows, execute `core-assemblies\apromore-core\target\assembly\bin\karaf.bat`
- When Apromore starts, it will present an interactive shell with a `karaf@root()>` prompt.  Beware that the server is still starting after the prompt appears.  Some of the logging isn't properly configured yet, so Liquibase in particular will disturb the console with misplaced log messages as it starts.  Pressing return will get the command prompt back.
- To see what's going on while the server is starting up, view the log file at `core-assemblies/apromore-core/target/assembly/data/log/karaf.log`.  The `tail -f` shell command or macOS Console utility are appropriate for this.
- Browse http://localhost:8181 as normal.  Apromore should behave indistinguishably on Karaf as it does on Virgo, other than the default port being different.
- Some commands to try from the shell:
  - `feature:list --installed | grep apromore` show the installed Apromore features (equivalent to Virgo plans)
  - `feature:info apromore-portal` show the details about a particular feature
  - `bundle:list` show the bundles
  - `bundle:find-class org.json.JSONObject` show all the bundles that use the JSONObject class
  - `package:exports -p com.google.common` show which bundles export the Guava packages
  - `http:list` show all the HTTP servlets
  - `web:list` show all the web applications (WARs)
  - `config:list "(service.pid=site)"` show the configured site properties
  - `config:property-set -p org.ops4j.pax.web org.osgi.service.http.port 9000` change the server's HTTP port
  - `bundle:services org.apromore.manager` show the manager's OSGi services
  - `service:list -n | grep org.apromore.plugin.portal` show the types and number of OSGi services using portal APIs
  - `log:display` show the log
  - `help` show all available shell commands
  - `system:shutdown` will stop the server, but Ctrl-D does the same thing more easily.
- The web console should be browsable at http://localhost:8181/system/console (username: karaf, password: karaf).  It includes a command shell at http://localhost:8181/system/console/gogo
- The shell can also be accessed via ssh to port 8101 with the same credentials as the web console, e.g. `ssh -p 8101 karaf@localhost` and password "karaf".
- The `site.properties` configuration file is located at `core-assemblies/apromore-core/target/assembly/etc/site.cfg`.
- The equivalent of the Virgo `pickup/` directory for dynamic deployment is `core-assemblies/apromore-core/target/assembly/deploy/`
  -  If you drop a non-bundle JAR into this directory, Karaf will attempt to turn it into a bundle on the fly by exporting every package in the JAR.  Convenient for quick-and-dirty development.

It's possible to run Apromore on Virgo port 9000 and Karaf port 8181 simultaneously.  However, you can't easily connect to both with the same browser since each time you log in to one, it will overwrite the ZK session cookie and log the other one out.  A workaround is to use two separate browsers.


.gitignore
```diff
@@ -25,6 +25,7 @@
 **/.DS_Store
 
 # Maven artifacts (i.e. object files)
+**/pom.xml.versionsBackup
```
raboczi, line 28: This is present because the `mvn versions:set` command creates backups of the pom.xml files it edits, but we don't want to commit those.

nolantellis (approved):
Tested on local with Vigro. Seems to work fine.

PR #633, opened 2020-12-24
AP-2807: add CaseLength filter

chiichang:


raboczi (approved):
Code looks fine style-wise, and I can build and start it without trouble.  I'm assuming you've confirmed the functional side of this; otherwise I approve it.

PR #644, opened 2021-01-04
AP-2919 Update copyright year in header

franklma:


raboczi (approved):
We might have trouble with merging, but I don't see any way to avoid it.

PR #648, opened 2021-01-04
AP-2899 Fix Access Right Mgmt left panel collapse

dreamind:


raboczi (approved):
Built, tried, left panel collapses as expected.

PR #652, opened 2021-01-05
AP-2793: fix inconsistency old and new animation

brucenguyen1:
This PR fixes the inconsistency between the old and new log animation logic which was caused by these errors:
1. Missing lifecycle:transition in the Optimizer, that's why many activities are skipped instead of played over.
Fix: add lifecycle:transition to the Optimizer.optimizeLog() method

2. Wrong cases of element mapping for animating on graphs
Fix: 
- add ElementIDMapper as a simple mapping solution
- improve ReplayTrace.convertToNonGateways() to use ElementIDMapper 
- improve LogAnimationServiceImpl2.transformToNonGateways() to use ElementIDMapper 

3. Improper error handling for wrong cases of element mapping (errors were not caught and reported)
Fix: add DiagramMappingException and throws this exception when it happens.

Note: these are quick fixes to the best extent for the legacy code (no unit tests) because the replay algorithm will be replaced in the next stage.

To test if this is fixed: follows card AP-2793, open PurchasingExample log in PD, click on the second play button (the new log animation), confirm that the "Amend Request for Quotation Requester" is actually a busy activity with many tokens flowing through, also confirm that most of activities on the model are actually played throught (the token flows through the center instead of on the edge).

Better use a git visual tool to view specific changes in these files because github is not smart enough to show these changes; it shows all file changes instead.

**Please use both Core and EE "feature/AP-2793-inconsistency-old-new-animation" branch.**

### DIFF
--- a/Apromore-Custom-Plugins/Log-Animation-Logic/src/main/java/org/apromore/service/loganimation/replay/Optimizer.java
+++ b/Apromore-Custom-Plugins/Log-Animation-Logic/src/main/java/org/apromore/service/loganimation/replay/Optimizer.java
@@ -29,6 +29,7 @@ import java.util.List;
 import java.util.concurrent.ConcurrentHashMap;
 
 import org.deckfour.xes.extension.std.XConceptExtension;
+import org.deckfour.xes.extension.std.XLifecycleExtension;
 import org.deckfour.xes.extension.std.XTimeExtension;
 import org.deckfour.xes.factory.XFactory;
 import org.deckfour.xes.factory.XFactoryNaiveImpl;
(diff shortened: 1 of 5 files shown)


raboczi (approved):
Code looks fine and runs fine.  I'm wondering whether GitHub's inability to do proper diffing has anything to do with the line-ending issues we had a while back when we were trying to duplicate the CSV Importer tests on your machine.

Apromore-Custom-Plugins/Log-Animation-Logic/src/main/java/org/apromore/service/loganimation/replay/Optimizer.java
```diff
@@ -1,166 +1,183 @@
...
+    }
+
+    private Object getAttributeValue(XAttribute attribute) {
+        if(attribute instanceof XAttributeLiteral) return ((XAttributeLiteral) attribute).getValue();
+        else if(attribute instanceof XAttributeBoolean) return ((XAttributeBoolean) attribute).getValue();
+        else if(attribute instanceof XAttributeDiscrete) return ((XAttributeDiscrete) attribute).getValue();
+        else if(attribute instanceof XAttributeContinuous) return ((XAttributeContinuous) attribute).getValue();
+        else if(attribute instanceof XAttributeTimestamp) return ((XAttributeTimestamp) attribute).getValue();
+        else {
+            System.out.println(attribute);
```
dreamind, line 164: Would it be better to use LOGGER?

dreamind (approved):

Apromore-Custom-Plugins/Log-Animation-Logic/src/main/java/org/apromore/service/loganimation/replay/Optimizer.java
raboczi: I was so distracted trying to figure out whether there was a way to avoid all that typecasting in the preceding lines (AFAICT there isn't) that I completely missed that!

brucenguyen1: Yeah, this section of code was written by the original author, I didn't look at it. It has some issues with the error handling is not proper  when the attribute is not one of the accepted types, only printing out with System.out.

The casting is one thing needs to do because of OpenXES's design, perhaps an issue, XAttribute doesn't have getValue() method in its interface, while getValue() is present in other extended interface such as XAttributeDiscrete, XAttributeTimestamp, etc.

brucenguyen1: We'll have to pay attention to these details in the new implementation.

PR #661, opened 2021-01-07
AP-2944 Update copyright year on About page

franklma:


raboczi (approved):
Changes don't get much more innocuous than this.

PR #662, opened 2021-01-07
AP-2942: Liquibase logging on Karaf

raboczi:
Someone else ran into the problem with Liquibase's logging not working on Karaf in 2017 and packaged a fix at https://github.com/steinarb/liquibase-karaf-feature/.  This PR applies the fix such that Liquibase starts cleanly and logs to the system log rather than the console.  Test it by executing `ApromoreCore/core-assemblies/apromore-core/target/assembly/bin/karaf` and observing that the console is no longer cluttered with spurious log messages.

I also changed the core PD and Log Animation ZK session cleanup hooks to log to SLF4J rather than System.out, as this was another source of misrouted log messages.

ApromoreEE inherits these changes automatically and requires no change.


raboczi:
The earliest version of Liquibase for which the fix is available is 3.5.5, which is what I used here.  More recent versions of Liquibase are supported, but they have more stringent validation and our current changesets will fail with an exception complaining that H2 doesn't support cluster operations.  We may want to revisit this in future and push the Liquibase version to at least 3.8.0.

nolantellis (commented):
I think liquibase-core was not a osgi module.
did you try it out running with a clean db on virgo ?

raboczi:
> I think liquibase-core was not a osgi module.
> did you try it out running with a clean db on virgo ?

Yes, I did.  Liquibase is haphazard: some versions of it are valid OSGi bundles, some of them aren't.  There's a list at https://github.com/steinarb/liquibase-karaf-feature/ of which versions are OSGi and which corresponding versions of snakeyaml are compatible.

nolantellis (approved):

PR #663, opened 2021-01-07
AP-2940

chiichang:
It seems most methods have been addressed in the other UnitTests written previously for testing the log filtering functions.
More UnitTest cases will be added in the future when issues are discovered.


chiichang:
getAttributes(XEvent xEvent) of LogFactory is a method that has been only used locally.

raboczi (approved):
The new tests run and pass, so approved.

The printString methods really ought to be removed because they're spammy for people doing builds and they're not even recorded in the test results (target/surefire-reports).  That's not an important change at the moment, though.

PR #666, opened 2021-01-07
AP-2917 Fix case variant percentage in case details

dreamind:


chiichang (approved):
No problem with build&run; Case variant percentage has been fixed.

PR #669, opened 2021-01-08
AP-2957 Modify logic to handle parent/ancestor folders' access rights…

franklma:
… when sharing

Change the parent/ancestor permission to Viewer unless they have been set higher access (Editor, Owner).


chiichang (approved):
- Build&run no problem
- Folder shared automatically with log

PR #674, opened 2021-01-08
AP-2447 Fix view 'keywords' in init script

franklma:
In line 1935 of changeLog-schema-init.yaml, an SQL view "keywords" is defined which has a column "value" which is a UNION of every field that might be searched.  What's happening is that some of those values are numerical: process.id and log.id, and that's confusing the view when it tries to merge the values into the one column.

### DIFF
--- a/Apromore-Database/src/main/resources/db/migration/changeLog-Schema.yaml
+++ b/Apromore-Database/src/main/resources/db/migration/changeLog-Schema.yaml
@@ -723,16 +723,6 @@ databaseChangeLog:
        - dropView:
            viewName: keywords
 
-
- - changeSet:
-     id: 20210108215600
-     author: frankm
-     preConditions:
-       - onFail: MARK_RAN
-       - not:
-           - viewExists:
-               viewName: keywords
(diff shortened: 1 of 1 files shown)


raboczi (commented):
I'm currently seeing test failures when I try to build apromore-database.

raboczi:
> I'm currently seeing test failures when I try to build apromore-database.

I spent a little more time tinkering with this.  If I comment out the preCondition clauses of the two new changeSet entries, then the tests pass.  This obviously isn't a solution, but maybe it's a clue for someone with a deeper knowledge of Liquibase to diagnose.

franklma:
> > I'm currently seeing test failures when I try to build apromore-database.
> 
> I spent a little more time tinkering with this. If I comment out the preCondition clauses of the two new changeSet entries, then the tests pass. This obviously isn't a solution, but maybe it's a clue for someone with a deeper knowledge of Liquibase to diagnose.
Thanks for your clue. Combining the two new changeSet entries can avoid this issue.

PR #675, opened 2021-01-08
AP-2447 Fix view 'keywords' in init script

franklma:
In line 1935 of changeLog-schema-init.yaml, an SQL view "keywords" is defined which has a column "value" which is a UNION of every field that might be searched. What's happening is that some of those values are numerical: process.id and log.id, and that's confusing the view when it tries to merge the values into the one column.


raboczi (commented):
I'm currently seeing test failures when I try to build apromore-database.

PR #680, opened 2021-01-09
AP-2472: implement multi-log animation

brucenguyen1:
This PR makes some changes to the animation logic to allow the new log animation portal plugin (EE) to animate multiple logs on a model (Portal > select multi logs and a model > select Analyze > Animate logs.) It also fixes some found bugs relating to BPMN model used in the animation.

Summary of Changes:

**Animation Logic (common for both old and new animation)**: this is mainly the replay algorithm.
- LogAnimationService2, LogAnimationServiceImpl2: change return type to AnimationResult for explicit type, return list of animation logs in the result, remove unused code, 
- AnimationJSONBuilder2: add frame index related JSON fields specific to each log and remove unused JSON fields (e.g. progress).
- AnimationLog: move ElementMapping from the logic to the portal plugin (renamed to ModelMapping) as it should be and it's common to multiple animation logs.
- BPMNDiagramHelper, ModelChecker: fix a bug relating to reading BPMN model. Because BPMN.io doesn't use gatewayDirection attribute like the Oryx BPMN java library, using this field leads to error for BPMN files saved in BPMN.io for animation (so far the animation is mostly run from PD or uses a BPMN model generated from PD which uses Oryx library, so this bug was not detected, but it will throw error after editing the model and save in BPMN.io). Remove the use of this field.
- Optimizer: fix a bug relating to BPMN model. BPMN.io doesn't use the "name" attribute of gateways (it is null) like the Oryx BPMN java library, must explicitly set it to empty string value to be used in the animation and avoid NPE.

**PD Plugin, Core build and plan files**
- Changes to remove the old animation menu item and the same button in PD. Note that the Log Animation Portal Plugin CE will be moved to CE repo after the new log animation has settled.

**This PR should be used at the same time with the same PR for the Log Animation Portal Plugin in ApromoreEE (if EE is used).**


raboczi (approved):
The only thing that surprises me is the disabling of the old log animation in Core, but this code is certainly good enough to approve.

Issue #182, opened 2019-01-19
Trying to install Apromore on Ubuntu, ant can't resolve dependencies


denoide1:
I've installed all the correct versions of the required packets such as maven 3.3.9, ant 1.9.2 and Server JRE

i can build the source code correctly with maven 3.3.9 but when i try to execute "ant start-virgo" 

ant logs show multiple errors regarding missing dependencies

Here's a sample of the error 

```
<HD0002E> Hot deploy failed for file '~metrics-editor-plugin-1.1.war'. org.eclipse.virgo.nano.deployer.api.core.DeploymentException: Dependency satisfaction failed
     [exec]     at org.eclipse.virgo.kernel.deployer.core.internal.PipelinedApplicationDeployer.driveInstallPipeline(PipelinedApplicationDeployer.java:362)
     [exec]     at org.eclipse.virgo.kernel.deployer.core.internal.PipelinedApplicationDeployer.doInstall(PipelinedApplicationDeployer.java:185)
     [exec]     at org.eclipse.virgo.kernel.deployer.core.internal.PipelinedApplicationDeployer.install(PipelinedApplicationDeployer.java:140)
     [exec]     at org.eclipse.virgo.kernel.deployer.core.internal.PipelinedApplicationDeployer.deploy(PipelinedApplicationDeployer.java:253)
     [exec]     at org.eclipse.virgo.nano.deployer.hot.HotDeploymentFileSystemListener.deploy(HotDeployerFileSystemListener.java:225)
     [exec]     at org.eclipse.virgo.nano.deployer.hot.HotDeploymentFileSystemListener.deployIfNotDeployed(HotDeployerFileSystemListener.java:237)
     [exec]     at org.eclipse.virgo.nano.deployer.hot.HotDeploymentFileSystemListener.onChange(HotDeployerFileSystemListener.java:88)
     [exec]     at org.eclipse.virgo.util.io.FileSystemChecker.notifyListeners(FileSystemChecker.java:380)
     [exec]     at org.eclipse.virgo.util.io.FileSystemChecker.check(FileSystemChecker.java:289)
     [exec]     at org.eclipse.virgo.nano.deployer.hot.WatchTask.run(WatchTask.java:49)
     [exec]     at java.lang.Thread.run(Thread.java:748)
[...]

PR #683, opened 2021-01-11
AP-2945: Karaf integration test

raboczi:
This PR has a counterpart https://github.com/apromore/ApromoreEE/pull/182 in ApromoreEE which must be merged simultaneously with it.

This PR provides integration tests in `ApromoreCore/core-assemblies/core-features-itest` which creates a stock Karaf server and confirms that various parts of Apromore Core can be installed from the Maven repo using shell commands and that any expected web applications (i.e. Apromore-Portal and REST-Endpoint) start successfully.

For most of the Karaf-related PRs I've tried to leave the Virgo assembly untouched, but in this case I've had to modify the default Virgo configuration to relocate `ehcache.xml` and `portalContext-security.xml` from `configuration/` to `etc/`.  This is because Karaf only allows configuration files to be installed within the `etc/` directory.  This change allows us to continue to use identical configuration files for both Virgo and Karaf.

Because the integration test creates an entire application server for a testing environment, it's very slow.  It takes 75 seconds to run on my machine.  Currently it's not run as part of the regular build.  To invoke it, you must build ApromoreCore first, and then enter `ApromoreCore/core-assemblies/core-features-itest/` and run `mvn clean install` there.


core-assemblies/core-features/src/main/feature/feature.xml
```diff
@@ -142,6 +142,7 @@
   <feature name="apromore-portal" version="${project.version}" description="Apromore core web browser interface">
     <feature>apromore-calendar</feature>
     <feature>apromore-manager</feature>
+    <feature>jms</feature>
```
raboczi, line 145: This fixes a bug the integration test picked up.  Previously the apromore-portal feature wouldn't start on its own due to a missing JMS API class.

nolantellis (approved):

PR #684, opened 2021-01-11
Replace view KEYWORDS selectQuery in LiquidBase

franklma:
In line 1935 of changeLog-schema-init.yaml, an SQL view "keywords" is defined which has a column "value" which is a UNION of every field that might be searched. What's happening is that some of those values are numerical: process.id and log.id, and that's confusing the view when it tries to merge the values into the one column.

A changeset is created to replace the selectQuery part of view KEYWORDS. It uses replaceIfExists attribute instead of preconditions to avoid failure of unit tests in Apromore-database.

### DIFF
--- a/Apromore-Database/src/main/resources/db/migration/changeLog-Schema.yaml
+++ b/Apromore-Database/src/main/resources/db/migration/changeLog-Schema.yaml
@@ -712,3 +712,46 @@ databaseChangeLog:
            referencedTableName: user
            validate: true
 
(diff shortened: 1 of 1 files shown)


nolantellis (commented):
Does this work with H2 ?

franklma:
> Does this work with H2 ?

Yes. Actually, the reason for this patch is to fix an H2-specific error.

nolantellis (approved):

PR #686, opened 2021-01-12
AP-2979

chiichang:
Unused old code caused performance issue. Removed.


raboczi (approved):
Rip it out!  Approved.

I presume there's going to be a corresponding PR for the development branch as well.

PR #688, opened 2021-01-12
Feature/multi storage/ap 2824

nolantellis:


Apromore-Core-Components/Apromore-Manager/src/main/java/org/apromore/service/impl/EventLogFileServiceImpl.java
```diff
@@ -34,41 +36,54 @@
...
+//            File file = new File(directoryPath + name);
+//            if (!file.exists()) file.createNewFile();
+//            outputStream = new FileOutputStream(file);
+//            serializer.serialize(log, outputStream);
+//            outputStream.close();
+//        } catch (Exception e) {
+//            e.printStackTrace();
+//            System.out.println("Error");
+//        }
+//    }
```
raboczi: What's the reasoning for holding on to this commented-out block?

Apromore-Core-Components/Apromore-Manager/src/main/java/org/apromore/service/impl/EventLogServiceImpl.java
```diff
@@ -191,171 +214,219 @@ public SummariesType readLogSummaries(Integer folderId, String searchExpression)
...
+	Group publicGroup = groupRepo.findPublicGroup();
+	if (publicGroup == null) {
+	    LOGGER.warn("No public group present in repository");
+	    return Collections.emptySet();
+	}
+
+	Set<GroupLog> publicGroupLogs = new HashSet<>(); /*
+							  * groupLogs .stream() .filter(groupLog ->
+							  * publicGroup.equals(groupLog.getGroup())) .collect(Collectors.toSet());
+							  */
```
raboczi: I can see that we want to retain this stream-based alternative implementation for the glorious day when we ditch Virgo and can use lambda expressions, but the way it's been reformatted is bizarre to me.

raboczi (approved):
This has been thoroughly scrutinized and reviewed by multiple people, so it's definitely time to merge and be damned.  Approved!

Apromore-Core-Components/Apromore-Manager/src/main/java/org/apromore/service/impl/EventLogFileServiceImpl.java
nolantellis: I will delete it.

Apromore-Core-Components/Apromore-Manager/src/main/java/org/apromore/service/impl/EventLogServiceImpl.java
nolantellis: Will delete this

raboczi: No, don't delete it.  We're holding on to it so that when we ditch Virgo we can replace the Java 7 loop with newfangled streaming code.  I was only commenting that reindenting it that way was weird (and might even give Checkstyle conniptions due to line length.)

PR #693, opened 2021-01-13
AP-2994 AP-3002

chiichang:
AP-2994 error caused by inheritance. 
AP-3002 caused by taking incorrect event indexes.
Corresponding UnitTests have been added.


raboczi (approved):
Tidy enough to merge.  (Although if we weren't rushing these hotfixes through, I'd poke you about for loops with manual indexing.)

PR #698, opened 2021-01-14
Update README.md

shefali316:


README.md
```diff
@@ -110,6 +91,31 @@ The default in a deployed server is that the `ehcache.xml` configuration file is
...
 
 
+### Backup and restore
+Apromore stores its data objects in two places:
+* Database: all data, except the event logs
+* Event logs which are by default located in the top-level `Event-Logs-Repository` directory
+
+As such, both need to be backed up and restored.
+* To backup h2 database, it is enough to copy across the `Manager-Repository.h2.db` file
+* To backup MySQL database, the following command may be used  (If prompted for password, enter the password of the ‘apromore’ user i.e ‘MAcri’):
```
nolantellis: The password depends on the USer suplied information. You might need to change this.

README.md
```diff
@@ -110,6 +91,31 @@ The default in a deployed server is that the `ehcache.xml` configuration file is
...
+```bash
+mysqldump -u apromore -p apromore > backup.sql
+```
+
+To backup only one table (rather than the whole database), the following command may be used:
+```bash
+mysqldump -u root -p apromore tablename > tablename.sql
```
nolantellis: remove root user

README.md
```diff
@@ -156,3 +162,31 @@ It can be configured to instead allow login based on an external LDAP directory.
   - `virgo-tomcat-server-3.6.4.RELEASE/configuration/login.conf`
 
 When the server starts with the reconfigured portal, it will automatically create new accounts for valid LDAP logins.
+
+
+## Change Port Number (optional)
+* Change line#12 in the `site.properties` file present in the `/ApromoreEE/ApromoreCore/Apromore-Assembly/virgo-tomcat-server-3.6.4.RELEASE/repository/usr` directory.
```
nolantellis: Try Avoiding mentioning line numbers in the doc, as they keep changing.
Be specific on the parameter name.

nolantellis (changes requested):

README.md
```diff
@@ -22,25 +22,6 @@ This document is relevant to both editions, but if you have checked out this Cor
   Note that version 8.0 is currently not supported.
```
raboczi: Chii reported last night that MySQL 5.6 no longer works, so we should remove it from the list and state that 5.7 is the only version supported.  (And doing this will count as completing https://apromore.atlassian.net/browse/AP-3015)

raboczi (changes requested):

README.md
```diff
@@ -156,3 +162,38 @@ It can be configured to instead allow login based on an external LDAP directory.
...
+
+> Out of memory while building.
+* Either invoke `mvn` as `mvn -Xmx1G -XX:MaxPermSize=256m` or set the system property `MAVEN_OPTS` to `-Xmx1G -XX:MaxPermSize=256m`
+
+> Server fails to start.
+* If either Apromore or PQL are configured to use MySQL, confirm that the database server is running.
+* If you already run another server (e.g. OS X Server) you may need to change the port number in `Supplements/Virgo/tomcat-server.xml`.
+
+> Login screen appears, but "admin" / "password" doesn't work.
+* You may need to run `ant create-h2` to populate the H2 database.
```
raboczi: Because we now use Liquibase, this problem of a non-initialized database should no longer occur.  (I'd suggest adding a Jira issue to remove the create-h2 target from build.xml as well, except that will happen automatically when all the Ant scripts are eliminated with the rest of the Eclipse Virgo assembly infrastructure.)

raboczi (commented):
I've made changed to reflect that ApromoreCore is now able to run independently of CE or EE.

raboczi (approved):

nolantellis (approved):

PR #703, opened 2021-01-17
AP-3024 Add more defensive prog for user admin

dreamind:


raboczi (approved):
Replacing a bunch of redundant assignments with a single new one looks like a win.

PR #710, opened 2021-01-21
AP-2990: Assorted build process improvements

raboczi:
This PR should be used along with its ApromoreEE counterpart https://github.com/apromore/ApromoreEE/pull/198.  It addresses various long-standing warning messages from Maven, as well as the original goal of simplifying the testing plugins.

- The version of the surefire Maven plugin which executes unit tests has been bumped.  Much unused configuration has been removed from it.
- The failsafe Maven plugin now handles integration tests, which have been renamed to match its default convention (*IT.java)
- The bundle plugin is configured from the top-level pom rather than inherited, avoiding warnings for non-bundle artifacts.
- The license plugin has been configured not to warn when there are no licensable files and had the ineffective failOnMissingHeader removed.
- The unused xeslite library has been removed.
- Various components have had the top-level pom.xml added as their parent so that they inherit settings which solve various warnings; in particular, the source character encoding.

Some issues left out of scope from this PR:
- Build warnings about OSGi private packages.  I'm leaving this until Virgo is removed and the bundle population is more well-behaved.
- Anything to do with parquet-osgi.  That thing is a bag of hurt and similarly not to be antagonised until Virgo goes away.
- Several tests aren't being run because they assume the default surefire naming convention (*Test.java) rather than (*UnitTest.java).  Once these tests are fixed, we should adopt the default naming convention to prevent further confusion.  Until then, I'm not activating any tests that weren't previously running.


pom.xml
```diff
@@ -1433,35 +1463,10 @@
 					<printSummary>true</printSummary>
 					<runOrder>alphabetical</runOrder>
 					<failIfNoTests>false</failIfNoTests>
-					<argLine>${argLine} ${surefire.jvm.args}</argLine>
 					<includes>
-						<include>**/*${unit.test.suffix}</include>
-						<include>**/*${endpoint.test.suffix}</include>
+						<include>**/*UnitTest.java</include>
```
mayankgituni, line 1467: I see that you are setting up the UnitTest convention here.

mayankgituni (approved):
Looks good!

PR #713, opened 2021-01-25
AP-3082: HTTP headers passed to event log REST endpoint support lists

raboczi:
Improvements to the `:rest-endpoint` component to accommodate the integration server.

- When an event log is POSTed to the manager REST endpoint, if it requires multiple values for a particular header (e.g. Apromore-Log-Ignored), both multiple headers and comma-delimited value lists are supported.  For example,
  ```
  Apromore-Case-Attribute: Nutrition
  Apromore-Case-Attribute: Taste
  ```
  is equivalent to
  ```
  Apromore-Case-Attribute: Nutrition, Taste
  ```

- Fixed doc comments in the REST endpoint; `mvn javadoc:javadoc` now successfully generates Javadoc

- The CSV Importer returns a more helpful error text if the passed headers are incorrect

### DIFF
--- a/Apromore-Custom-Plugins/REST-Endpoint/src/main/java/org/apromore/rest/manager/ArtifactResource.java
+++ b/Apromore-Custom-Plugins/REST-Endpoint/src/main/java/org/apromore/rest/manager/ArtifactResource.java
@@ -29,6 +29,7 @@ import java.io.FileInputStream;
 import java.io.InputStream;
 import java.nio.file.Files;
 import static java.nio.file.StandardCopyOption.REPLACE_EXISTING;
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.GregorianCalendar;
 import java.util.HashMap;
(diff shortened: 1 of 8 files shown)


Apromore-Custom-Plugins/REST-Endpoint/src/main/java/org/apromore/rest/manager/ArtifactResource.java
```diff
@@ -318,10 +322,23 @@ private String findAttribute(final String name) {
...
+    /**
+     * @param name  an HTTP header name
+     * @return all values of the <var>name</var>d header occurring within {@link #httpHeaders}, honoring both
+     *     duplicate headers and headers with comma-delimited lists of values
+     */
     private List<String> findAttributes(final String name) {
         List<String> requestHeaders = httpHeaders.getRequestHeader(name);
 
-        return requestHeaders == null ? Collections.emptyList() : requestHeaders;
+        if (requestHeaders == null) {
```
karthik-balasubramanian, line 333: Seems to be redundant check

karthik-balasubramanian (approved):

Apromore-Custom-Plugins/REST-Endpoint/src/main/java/org/apromore/rest/manager/ArtifactResource.java
raboczi: The getRequestHeader method returns null rather than an empty list, so the check is necessary to guarantee the method always returns a List.

PR #717, opened 2021-01-27
AP-3085 Fix recursive copy and paste

dreamind:


raboczi (approved):
It's one of those "blindingly obvious" things, innit?

PR #724, opened 2021-01-29
Content-type for javascript files in the Portal (AP-3102) and PD message for logs with excessively large number of activities (AP-3083)

brucenguyen1:
This PR fixes the issues raised in AP-3083 and AP-3102.

AP-3083: raises message when the activities (concept:name) exceed the maximum allowed number of distinct values: 
Changes: modified config data from 300 to 500 (new limit), modified LogData to read the activity attribute, modified PDController to check and raise the message.

AP-3102: Log Animation EE can't run on Firefox because Firefox prevents it from loading web worker javascript. The reason is because the ResourceServlet for loading server resources didn't set a content type for javascript (text/javascript). So far, these files are loaded with empty content-type value but Firefox enforces stricter security rules (https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types#JavaScript_types).

Testing: just follows the description in AP-3083 and AP-3102. 
For AP-3083: try opening a log with more than 500 distinct activities, a message should display and PD stops there without proceeding to avoid server crash.
For AP-3102: in the current setting, log animation can't run on Firefox, but with the explicitly set content-type, it will run normally.

### DIFF
--- a/Apromore-Core-Components/Apromore-Portal/src/main/java/org/apromore/portal/servlet/ResourceServlet.java
+++ b/Apromore-Core-Components/Apromore-Portal/src/main/java/org/apromore/portal/servlet/ResourceServlet.java
@@ -22,24 +22,26 @@
 
 package org.apromore.portal.servlet;
 
-import com.google.common.io.ByteStreams;
-import java.io.InputStream;
 import java.io.IOException;
+import java.io.InputStream;
 import java.io.OutputStream;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+
 import javax.servlet.ServletException;
-import javax.servlet.ServletRegistration;
 import javax.servlet.http.HttpServlet;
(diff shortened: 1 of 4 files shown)


nolantellis (approved):

dreamind (approved):

Apromore-Core-Components/Apromore-Portal/src/main/java/org/apromore/portal/servlet/ResourceServlet.java
```diff
@@ -68,6 +70,7 @@ public void init() throws ServletException {
         contentTypeMap.put("svg", "image/svg+xml"); // for customised icon-fonts
         contentTypeMap.put("ttf", "application/font-sfnt"); // for customised icon-fonts
         contentTypeMap.put("woff", "application/font-woff"); // for customised icon-fonts
+        contentTypeMap.put("js", "text/javascript");
     }
```
dreamind, line 74: While you are at, probably we can include jpg and json.

PR #726, opened 2021-01-31
AP-2570 Restructure access control

dreamind:


Apromore-Core-Components/Apromore-BPMNEditor/src/main/resources/bpmneditor/WEB-INF/xml/editor/plugins.xml
```diff
@@ -33,6 +33,7 @@
         <plugin source="export.js" name="ORYX.Plugins.Export"/>
```
nolantellis: There is a lot to review. You might need to take a walk through.

franklma (dismissed):

franklma (approved):
Confirmed that ApromoreCore builds and passes all the tests. Made some minor changes.

PR #733, opened 2021-02-02
Fixes for Windows compile (and multi-platform)

brucenguyen1:
Apromore-Storage: changes URL.getFile() to URL.toURI() as the former results in strange character as path separater on Windows.
Surefire plugin configuration in the top pom file: add fork and memory settings. Without these settings, Apromore-Database unit tests often result in VM crashes (out of memory), sometimes also happened to APMLogModule.


Apromore-Core-Components/apromore-storage/src/test/java/org/apromore/test/storage/StorageFactoryUnitTest.java
```diff
@@ -46,11 +46,12 @@ public void setup()
...
     
     @Test
-    public void testFileStorageFactory() throws ObjectNotFoundException
+    public void testFileStorageFactory() throws ObjectNotFoundException, URISyntaxException
     {
 //	Given
 	ClassLoader classLoader = StorageFactoryUnitTest.class.getClassLoader();	
-	File file = new File(classLoader.getResource("baseFolder"+File.separator+"test.csv").getFile());
+	//File file = new File(classLoader.getResource("baseFolder"+File.separator+"test.csv").getFile());
+	File file = new File(classLoader.getResource("baseFolder"+File.separator+"test.csv").toURI());
```
raboczi: I don't think it's necessary to use the system-specific File.separator because this is a Java classpath rather than a path on the native filesystem.  Could you check if it still works on Windows to simply use "baseFolder/test.csv"?

pom.xml
```diff
@@ -1463,6 +1463,9 @@
 					<printSummary>true</printSummary>
 					<runOrder>alphabetical</runOrder>
 					<failIfNoTests>false</failIfNoTests>
+					<forkCount>3</forkCount>
+                    <reuseForks>true</reuseForks>
+                    <argLine>-Xmx1024m</argLine>
```
raboczi: The indentation here suggests that lines 1467 and 1468 are indented with spaces, whereas the rest of the file is indented with tabs.  I despise tabs (because they cause exactly this sort of problem) but for consistency the spaces here should be swapped for tabs.

raboczi (approved):
This all builds and tests successfully on my machine as well.  I've suggested some changes, but nothing sufficiently critical to block merging.

Apromore-Core-Components/apromore-storage/src/test/java/org/apromore/test/storage/StorageFactoryUnitTest.java
brucenguyen1: Sure it will work because no difference. But File.separator is usually recommended as a safe OS-independent although recent Java versions can do the same for the "/" character. However, the classpath on Windows and Unix still have differences.

pom.xml
brucenguyen1: don't know why, suppose my Eclipse has been set to replace all tab with spaces. Let me check.

raboczi: Maybe we should consider reformatting the entire pom.xml to remove the tabs.  It'd be spammy, but it has to be done eventually.  I think Nolan's IDE was set to prefer tabs at one point and that's what reformatted several files this way.

Apromore-Core-Components/apromore-storage/src/test/java/org/apromore/test/storage/StorageFactoryUnitTest.java
brucenguyen1: Actually you're right Simon. File.separator is the root cause of this issue. On Windows, File.separator will be converted to backslash "\\" and then this character in the file path will be converted to "%5c", so the file path will contain "baseFolder%5ctest.csv" which was the problem of File not found. Using forward-slash "baseFolder/test.csv" and Java will convert it to a correct system-specific format, on Windows it is "baseFolder\\\\test.csv". I think this is what happens in the classloader. Call URL.toURI() can fix the strange character "%5c" in the path but it is not the root cause.

raboczi: That looks good now.  👍

pom.xml
dreamind: 4-space indent or 2-space indent? :)

raboczi: I'm a two-space guy insofar as XML goes, but four would be acceptable.  As far as I recall, Checkstyle can't enforce 2-vs-4 space indenting, but it definitely can ban tabs from occurring in source code.  ✊

PR #745, opened 2021-02-05
Added Hikari pooling and removed BonCp

nolantellis:
Removed BonCp and added hikari
* Removed not required modules of plugin-database


raboczi:
I found some additional dependencies on the now-removed database driver fragments and removed them.

raboczi (approved):
I tested with MySQL 8 as well successfully.  Less code is best code!

PR #751, opened 2021-02-08
sync update from v7.19.1

chiichang:


raboczi (approved):
This is very large PR with too much material for a really close review, but I've confirmed it builds, confirmed that the changes are restricted to APMLog/Dashboard/FilterEE, and given a cursory skim that nothing looks outrageous.  I'll assume you'll vouch for the functional behavior and trust that the reviews of the preceding 7.19 PRs apply to this as well.

PR #755, opened 2021-02-08
AP-3222: Improve ResourceServlet to avoid crashes caused by misbehaved plugins

brucenguyen1:
Change ResourceServlet:
- Remove duplicate code
- Improve processing flow
- Add exception handling to handle exceptions thrown from plugins: the purpose is keeping the resource loading continue in case of plugin errors.

Testing: can turn a current plugin into a "bad" one to test with different scenarios: 
- Wrong service property key (configured in context.xml)
- Runtime exception in hasResource or getResourceAsStream() methods: e.g. NPE or IndexOutOfBoundException
- Missing resource
Check that the Portal shows UI normally (icon, images, styles, etc.) without stopping or losing styles. The console should display error message captured from exception handling as shown in the screenshot.
image

### DIFF
--- a/Apromore-Core-Components/Apromore-Portal/src/main/java/org/apromore/portal/servlet/ResourceServlet.java
+++ b/Apromore-Core-Components/Apromore-Portal/src/main/java/org/apromore/portal/servlet/ResourceServlet.java
@@ -22,24 +22,28 @@
 
 package org.apromore.portal.servlet;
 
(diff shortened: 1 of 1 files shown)


nolantellis (approved):

raboczi (approved):
Corrected PORTAL_SERVLET_BUNLDE_KEY typo, made it a static constant, and included the exceptions in a few log messages.

PR #763, opened 2021-02-09
AP-3233 Fix method only return Logs in root folder

franklma:


raboczi (approved):
That definitely wasn't doing what it advertised; must've been a real headache to find.  :P

PR #765, opened 2021-02-09
AP-3224 Fix shortenName method

dreamind:


raboczi (approved):
Rewrote the unit test to use JUnit's assertEquals method.

PR #770, opened 2021-02-12
Consistent version for maven-resources-plugin

raboczi:
The :core-features target was previously failing because it was being built with version 2.7 of the maven-resources-plugin.  Changed the system-wide version to 3.2.0.

Cleaned up a number of redundant similar declarations of plugin versions from the top-level pom.xml's pluginManagement section so that nobody will get confused by them in future.


franklma:
Build & Run without issue and make pom file neater.

franklma (approved):

PR #771, opened 2021-02-13
AP-2870: JDK 11 compatibility

raboczi:
This PR makes it possible under JDK 11 for Apromore to build (albeit with tests skipped) and run.  Only Karaf can run on JDK 9+; Virgo does not even fully support JDK 8.

The project still treats the source code as Java 8 and targets a Java 8 JVM, so theoretically the bundles are unchanged. The difference is that a JDK 11 environment won't contain the various javax.* packages which were dropped from Java SE after version 11. By adding a profile to the top-level pom.xml which adds these packages as explicit additional dependencies when Maven detects a JDK 11+ environment, the build will continue to work.

While this change makes building and running under JDK 11 possible, it only does so with tests suppressed due to the following test failures:
- Apromore-Calendar fails due to the modelmapper library
- CSVImporter-Logic shows an error reading hebrew CSV files

I'd argue in favor merging even with the test failures present.  This PR does not affect operation on JDK 8, and having it as part of the development branch prevents debugging them from being a race against bit rot.

The ApromoreEE repository inherits these changes from ApromoreCore and should build on JDK 11 without itself requiring any change, but the Karaf assemblies will require their own profiles.   There will eventually be a counterpart PR.

### DIFF
--- a/Apromore-Core-Components/Apromore-Portal/pom.xml
+++ b/Apromore-Core-Components/Apromore-Portal/pom.xml
@@ -69,6 +69,7 @@
                         <Import-Package>
                             org.eclipse.virgo.web.dm;resolution:=optional,
                             org.aopalliance.aop,                            
+                            org.apromore.security.exception,
                             org.apromore.security.filter,
                             org.apromore.security.provider,
                             org.apromore.service,
--- a/Apromore-Extras/Chiaroscuro-Theme/pom.xml
+++ b/Apromore-Extras/Chiaroscuro-Theme/pom.xml
@@ -122,6 +122,7 @@
 				</executions>
 			</plugin>
 			<!-- Compile java -->
+<!--
 			<plugin>
 				<groupId>org.apache.maven.plugins</groupId>
 				<artifactId>maven-compiler-plugin</artifactId>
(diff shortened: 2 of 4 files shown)


Apromore-Core-Components/Apromore-Portal/pom.xml
```diff
@@ -69,6 +69,7 @@
                         <Import-Package>
                             org.eclipse.virgo.web.dm;resolution:=optional,
                             org.aopalliance.aop,                            
+                            org.apromore.security.exception,
```
raboczi, line 72: This is a bugfix not directly related to JDK 11 compatibility.  The missing package only causes trouble when an exception is thrown, which is why it wasn't noticed until I had to debug this PR.

Apromore-Extras/Chiaroscuro-Theme/pom.xml
```diff
@@ -121,20 +121,6 @@
...
 				</executions>
 			</plugin>
-			<!-- Compile java -->
-			<plugin>
-				<groupId>org.apache.maven.plugins</groupId>
-				<artifactId>maven-compiler-plugin</artifactId>
-				<version>2.3.2</version>
-				<configuration>
-					<source>1.6</source>
-					<target>1.6</target>
```
raboczi, line 131: Later JDKs no longer support targetting these very early versions.  Removing this block allows the compiler version to be specified by the top-level ApromoreCore/pom.xml instead.

nolantellis (approved):

PR #778, opened 2021-02-16
AP-3256 Change rules on validating timestamp in CSV Importer

franklma:
A valid timestamp must have the date, month and year present.


raboczi (approved):
Confirmed that ApromoreCore builds, passes the tweaked tests, runs, and CSV importing still works.  The code changes seem tidy and (for once) have tests.  :)

PR #779, opened 2021-02-16
AP-3258 Log with invalid cells uploaded twice

franklma:
Hotfix of AP-3258 Log with invalid cells uploaded twice (https://apromore.atlassian.net/browse/AP-3258)
This is a regression introduced by the major refactoring last year.

### DIFF
--- a/Apromore-Custom-Plugins/CSVImporter-Portal/src/main/java/org/apromore/plugin/portal/csvimporter/CSVImporterController.java
+++ b/Apromore-Custom-Plugins/CSVImporter-Portal/src/main/java/org/apromore/plugin/portal/csvimporter/CSVImporterController.java
@@ -1,7 +1,7 @@
 /*-
  * #%L
  * This file is part of "Apromore Core".
- * 
+ *
  * Copyright (C) 2020 University of Tartu
  * %%
  * Copyright (C) 2018 - 2021 Apromore Pty Ltd.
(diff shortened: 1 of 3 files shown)


Apromore-Custom-Plugins/CSVImporter-Portal/src/main/java/org/apromore/plugin/portal/csvimporter/CSVImporterController.java
```diff
@@ -1110,6 +1065,10 @@ private void saveXLog(LogModel logModel, boolean isPublic) {
                 portalContext.refreshContent();
 
             } else {
+
+                if (logModel.getImportLog() == null) {
+                }
+
```
raboczi: Did the body of this conditional miss getting filled in?

raboczi (changes requested):
I've confirmed it builds and runs.  There's only one bit of suspicious code that looks like something is missing.  Otherwise I'll be happy to approve this.

Apromore-Custom-Plugins/CSVImporter-Portal/src/main/java/org/apromore/plugin/portal/csvimporter/CSVImporterController.java
franklma: This can be safely removed.

raboczi (approved):

PR #781, opened 2021-02-19
Task/calendar log/ap 2814

nolantellis:


Apromore-Core-Components/Apromore-Manager/src/main/java/org/apromore/service/EventLogService.java
```diff
@@ -106,4 +106,9 @@ void cloneLog(String username, Integer folderId, String logName, Integer sourcel
      * @return The aggregated log placed into the cache, or generated on the fly if not found or expired
      */
     APMLog getAggregatedLog(Integer logId);
+
+
+    void updateCalenderForLog(Integer logId, Long calenderId);
```
raboczi: Should be "updateCalendarForLog"

Apromore-Core-Components/Apromore-Manager/src/main/java/org/apromore/service/impl/EventLogServiceImpl.java
```diff
@@ -446,4 +433,20 @@ public APMLog getAggregatedLog(Integer logId) {
 	return tempCacheService.getAggregatedLog(log);
     }
 
+
+    @Override
+    public void updateCalenderForLog(Integer logId, Long calenderId) {
```
raboczi: Should be "updateCalenderForLog"

raboczi (changes requested):
Builds and runs on both the Virgo and Karaf versions of Core, code seems tidy.  Just a trivial typo to fix.

raboczi (approved):

PR #782, opened 2021-02-19
AP-3237: Modified CSVImporterController for extension

raboczi:
This PR is required to support https://github.com/apromore/ApromoreEE/pull/233 in ApromoreEE.

 Changes to make CSVImporterController extensible.
- Resources obtained from the classpath now use `CSVImporterController.class.getClassLoader()` rather than `getClass().getClassLoader()` so that subclasses can still use them
- The `.getLogModel()` method was factored out


mayankgituni (approved):
Works!!

PR #785, opened 2021-02-23
AP-3260: Remove Virgo

raboczi:
This PR has a counterpart https://github.com/apromore/ApromoreEE/pull/238 in ApromoreEE which should be checked out and merged simultaneously.

Remove the Eclipse Virgo assembly and change the OSGi version from 4.2 to 7.0.


franklma (approved):
Build and deploy without issue. Run a quick test, everything works as expected.

PR #787, opened 2021-02-23
AP-3244 Refine undo/redo for PD

dreamind:


brucenguyen1:
As discussed, a server-side action management would be more suitable and extensible than a front-end approach. This is because many actions in the system are created with server code (Java) rather than front-end code (javascript) and these Java data structures can be reused in action management instead of recreating them in Javascript.

dreamind:
@brucenguyen1 I push an update as a fix for the filter by case detail before the better solution is implemented.

brucenguyen1:
We'll address this PR more completely in the next release.

brucenguyen1:
We'll address this PR more completely in the next release to cover all filter actions.

PR #788, opened 2021-02-24
AP-3317

chiichang:
Fixed Arc Duration Filter logic and added corresponding UnitTest


Apromore-Extras/APMLogModule/src/main/java/org/apromore/apmlog/filter/typefilters/AttributeArcDurationFilter.java
```diff
@@ -74,7 +76,9 @@ private static boolean conformRule(ATrace trace, LogFilterRule logFilterRule) {
                                                                     double lowBoundVal, double upBoundVal) {
         List<Double> durList= new ArrayList<>();
 
-        List<AActivity> activityList = trace.getActivityList();
+        List<AActivity> activityList = trace.getActivityList().stream()
+                .filter(x -> trace.getValidEventIndexBitSet().get(x.getEventIndexes().get(0)))
```
raboczi, line 80: Is there a guarantee that `trace.getActivityList().getEventIndexes()` will never be empty?  Otherwise the `.get(0)` call in this filter could fail.

raboczi (approved):

PR #789, opened 2021-02-24
AP-3317

chiichang:
Counterpart of AP-3317-v7.19 [PR#788](https://github.com/apromore/ApromoreCore/pull/788)

### DIFF
--- a/Apromore-Extras/APMLogModule/src/main/java/org/apromore/apmlog/filter/typefilters/AttributeArcDurationFilter.java
+++ b/Apromore-Extras/APMLogModule/src/main/java/org/apromore/apmlog/filter/typefilters/AttributeArcDurationFilter.java
@@ -32,6 +32,7 @@ import org.eclipse.collections.impl.map.mutable.UnifiedMap;
 
 import java.util.ArrayList;
 import java.util.List;
+import java.util.stream.Collectors;
 
 public class AttributeArcDurationFilter {
     public static boolean toKeep(ATrace trace, LogFilterRule logFilterRule) {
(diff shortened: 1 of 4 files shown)


Apromore-Extras/APMLogModule/src/main/java/org/apromore/apmlog/filter/typefilters/AttributeArcDurationFilter.java
```diff
@@ -74,7 +75,9 @@ private static boolean conformRule(ATrace trace, LogFilterRule logFilterRule) {
                                                                     double lowBoundVal, double upBoundVal) {
         List<Double> durList= new ArrayList<>();
 
-        List<AActivity> activityList = trace.getActivityList();
+        List<AActivity> activityList = trace.getActivityList().stream()
+                .filter(x -> trace.getValidEventIndexBitSet().get(x.getEventIndexes().get(0)))
```
raboczi, line 79: Similarly to the 7.19 version of this PR, is it possible for this code to fail if one of the trace activities returned by `trace.getActivityList()` has an empty list returned by `.getEventIndexes()`?

raboczi (approved):

Apromore-Extras/APMLogModule/src/main/java/org/apromore/apmlog/filter/typefilters/AttributeArcDurationFilter.java
chiichang: If the trace has empty events/activities, it wouldn't be included in the log.

PR #794, opened 2021-03-01
AP-2702: Remove/consolidate dependencies

raboczi:
This PR is paired with https://github.com/apromore/ApromoreEE/pull/250 in ApromoreEE.  They need to be checked out, built, and merged concurrently.

Cleanup of pom.xml files:
* consolidate around a single version of the JAXB and JSON libraries, both as freestanding bundles and as embeddings within other bundles
* moved version declarations for lombok to be shared from ApromoreCore/pom.xml
* removed logging libraries other than the SLF4J API
* general cleanup of dead code from the manager and portal pom.xml

### DIFF
--- a/Apromore-Calendar/pom.xml
+++ b/Apromore-Calendar/pom.xml
@@ -52,8 +52,6 @@
 		<dependency>
 			<groupId>org.projectlombok</groupId>
 			<artifactId>lombok</artifactId>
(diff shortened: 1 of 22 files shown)


franklma (approved):
It must be a really satisfying clean-up : )

Build and deploy successfully.

PR #800, opened 2021-03-04
Changed default DB to SQL

shefali316:


raboczi (approved):
Straightforward enough; approved.

PR #801, opened 2021-03-04
Change default db to SQL

shefali316:


raboczi (approved):
This seems in order; approved.

PR #805, opened 2021-03-09
AP-3360: Unable to animate log after importing from CSV file

brucenguyen1:
Log Animation Logic reads the "concept:name" attribute of XLog (representing log name). This was implemented many years ago. Recent changes in the CSVImporter has removed this attribute from the cached XLog created after importing a CSV file which caused NPE in Log Animation Logic.

This fix makes changes to Log Animation Logic to read the log name from the selected LogSummary item on the portal instead of reading the "concept:name" attribute in XLog. Meanwhile, the color attribute is now managed by the front-end (Javascript) instead of backend, so it is removed in this PR.

This bug only occurs on release/v7.19 branch, not on development. However, it should also be applied to v7.20 to avoid any potential similar changes in CSV Importer.

To verify this fix: import a CSV file into a log item, open that log in PD, click "Animate Log", confirm the log animation window is opened and the log can be animated.

### DIFF
--- a/Apromore-Custom-Plugins/Log-Animation-Logic/src/main/java/org/apromore/service/loganimation/replay/Replayer.java
+++ b/Apromore-Custom-Plugins/Log-Animation-Logic/src/main/java/org/apromore/service/loganimation/replay/Replayer.java
@@ -101,10 +101,10 @@ public class Replayer {
         return processCheckMessage;
     }
     
-    public AnimationLog replay(XLog log, String color) {
-        AnimationLog animationLog = new AnimationLog(log);
-        animationLog.setColor(color /*this.getLogColor()*/);
-        animationLog.setName(log.getAttributes().get("concept:name").toString());
+    public AnimationLog replay(XLog log, String name) {
+        AnimationLog animationLog = new AnimationLog(log, name);
+        //animationLog.setColor(color /*this.getLogColor()*/);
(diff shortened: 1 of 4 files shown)


Apromore-Custom-Plugins/Log-Animation-Logic/src/main/java/org/apromore/service/loganimation/replay/Replayer.java
```diff
@@ -101,10 +101,10 @@ public String getProcessCheckingMsg() {
...
     }
     
-    public AnimationLog replay(XLog log, String color) {
-        AnimationLog animationLog = new AnimationLog(log);
-        animationLog.setColor(color /*this.getLogColor()*/);
-        animationLog.setName(log.getAttributes().get("concept:name").toString());
+    public AnimationLog replay(XLog log, String name) {
+        AnimationLog animationLog = new AnimationLog(log, name);
+        //animationLog.setColor(color /*this.getLogColor()*/);
+        //animationLog.setName(log.getAttributes().get("concept:name").toString());
```
raboczi, line 107: Unless there's reason to believe these commented-out lines will ever return, may as well delete them for good.

raboczi (approved):
Confirmed the defect without this PR, and that the PR does indeed address it.  Changes are tidy and safe to merge.  I only nitpicked about some commented-out code that could probably do with removal.

PR #814, opened 2021-03-14
Feature/ap 3145/job scheduler

Janite:
This has a corresponding PR in ApromoreEE https://github.com/apromore/ApromoreEE/pull/263


Apromore-Custom-Plugins/CSVImporter-Portal/src/main/java/org/apromore/plugin/portal/csvimporter/CSVImporterController.java
```diff
@@ -393,10 +393,25 @@ private void storeMappingAsJSON(Media media, LogMetaData logMetaData, Log log) t
                 log.getId());
     }
 
+    protected String getLogTag() {
```
nolantellis, line 396: This is not the right place. This does not belong to a controller.

site.properties
```diff
@@ -39,13 +39,13 @@ etlplugin.pipeline.s3.region = ap-southeast-2
 airflow.username = airflow
 airflow.password = airflow
 airflow.hostname = http://localhost:8080
+etl.s3.url = S3::mayank-test-databucket::ap-southeast-2::https://s3.ap-southeast-2.amazonaws.com
```
nolantellis: Is there a chance , we can chnage this to a File location as default. 
We can chat for more context.

nolantellis (commented):
Rest is fine.

nolantellis (approved):

PR #817, opened 2021-03-16
AP-3395 Access rights propagation

franklma:
Access control lists for a folder propagate downward and are inherited by all child files and folders. Whenever ACLs or the hierarchy is changed, the propagation occurs recursively through all nested folders. For example, if a file exists in a folder and that folder is then moved within another folder, the ACLs on the new folder propagate to the file. If the new folder grants the user of the file a new access type, such as "editor," it overrides their old role. Conversely, if a file inherits the "editor" role from a folder, and is moved to another folder that provides a "viewer" role, the file now inherits the "viewer" role.

Inherited ACLs can be overridden on a file or folder. So, if a file inherits the role of "editor" from a folder, you can set the role of "reader" on the file to lower its permission level.

### DIFF
--- a/Apromore-Core-Components/Apromore-Manager/src/main/java/org/apromore/service/impl/FolderServiceImpl.java
+++ b/Apromore-Core-Components/Apromore-Manager/src/main/java/org/apromore/service/impl/FolderServiceImpl.java
@@ -171,7 +171,7 @@ public class FolderServiceImpl implements FolderService {
         String chain = oldFolder.getParentFolderChain() + "_" + id;
         String prefix = chain + "_";
         prefix = getEscapedString(prefix) + "%";
-        List<Folder> folders = folderRepository.findByParentFolderIdOrParentFolderChainLike(id, prefix);
+        List<Folder> folders = new ArrayList(folderRepository.findByParentFolderIdOrParentFolderChainLike(id, prefix));
         if (includeCurrentFolder) {
             folders.add(oldFolder);
         }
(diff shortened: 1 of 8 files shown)


Apromore-Core-Components/Apromore-Manager/src/main/java/org/apromore/service/impl/FolderServiceImpl.java
```diff
@@ -171,7 +171,7 @@ private String getEscapedString(String prefix) {
         String chain = oldFolder.getParentFolderChain() + "_" + id;
         String prefix = chain + "_";
         prefix = getEscapedString(prefix) + "%";
-        List<Folder> folders = new ArrayList(folderRepository.findByParentFolderIdOrParentFolderChainLike(id, prefix));
```
nolantellis: you will need to move it back to new ArrayList for jpa reasons.

Apromore-Core-Components/Apromore-Manager/src/main/java/org/apromore/service/impl/WorkspaceServiceImpl.java
```diff
@@ -651,6 +711,18 @@ public Process moveProcess(Integer processId, Integer newFolderId) throws Except
         Folder newFolder = folderRepo.findUniqueByID(newFolderId);
         Process process = processRepo.findUniqueByID(processId);
         process.setFolder(newFolder);
+
+        // Unless in the root folder, overwrite and inherit access rights from direct parent folder
+        if (!newFolderId.equals(ROOT_FOLDER_ID)) {
```
nolantellis: reverse the variables  !ROOT_FOLDER_ID.equals(newFolderId)

nolantellis (approved):

PR #837, opened 2021-03-24
AP-2537 Timeout for deleted user session

franklma:


chiichang (approved):

Apromore-Core-Components/Apromore-Portal/src/main/java/org/apromore/portal/dialogController/UserMenuController.java
```diff
@@ -186,6 +186,20 @@ public void onEvent(Event event) {
...
+                        public void onEvent(Event event) {
+                            Session session = Sessions.getCurrent();
+
+                            UserType userType = (UserType) Sessions.getCurrent().getAttribute("USER");
+                            if (session == null || event.getData().equals(userType.getUsername())) {
+                                Executions.sendRedirect("/j_spring_security_logout");
+                            }
+                        }
+                    }
+            );
```
raboczi, line 202: Two possible improvements: abbreviate using a lambda expression, and handle the case of a null USER attribute.
```
import java.util.Objects;
```
```

// Force logout a user that has been deleted by admin
EventQueues.lookup("forceSignOutQueue",  EventQueues.APPLICATION, true).subscribe(event -> {
    Session session = Sessions.getCurrent();
    if (session != null) {
        UserType userType = (UserType) session.getAttribute("USER");
        if (userType != null && event != null && Objects.equals(event.getData(), userType.getUsername())) {
            // The user's session is still valid, so no action required
            return;  
        }
    }   

    // The user's session has indeed been invalidated, so force logout
    Executions.sendRedirect("/j_spring_security_logout");
});
```

raboczi (commented):
I have a suggested improvement, but otherwise this seems safe to merge.

PR #840, opened 2021-03-28
feat: added delete and copy method.

mayankgituni:
This PR must be paired with reverting https://github.com/apromore/ApromoreEE/pull/293 such that the original counterpart PR in ApromoreEE https://github.com/apromore/ApromoreEE/pull/271 is reinstated.


site.properties
```diff
@@ -44,25 +44,31 @@ mail.username = user@gmail.com
...
+airflow.hostname = http://localhost:8080
+etl.s3.url = S3::mayank-test-databucket::ap-southeast-2::https://s3.ap-southeast-2.amazonaws.com
+apromore.tenant = ap
+
+# Impala
+impala.jdbc.url      = jdbc:impala://localhost:21050/
+impala.jdbc.username = apromore
+impala.jdbc.password = MAcri
+impala.etl.dataStore = /etl_data
+impala.etl.s3.directory = users/
```
raboczi: These properties don't look like they belong in the open source default configuration.

raboczi (changes requested):
Probably shouldn't've committed site.properties at all.

raboczi (approved):
This now builds if the companion PR for ApromoreEE is also applied.

PR #843, opened 2021-03-29
AP-3471 Report issue menu

dreamind:


Apromore-Core-Components/Apromore-Portal/src/main/webapp/themes/ap/common/js/index.js
```diff
@@ -113,4 +113,21 @@ Ap.common.pullClientTimeZone = function () {
...
+ * @param email {String} Email
+ * @param username {String} Username
+ */
+Ap.common.reportIssue = function (email, username) {
+    let subject = 'Apromore Issue Report';
+    let ua = (typeof navigator === 'undefined') ? '' : navigator.userAgent.toLowerCase();
+    let href = window.location.href;
+    let body = `Username: ${username}%0D%0A` + `Browser info: ${ua}%0D%0A` + `URL: ${href}%0D%0A`;
+    body += 'Description of your issue:'
+    let url = "mailto:" + email + "?subject=" + subject + "&body=" + body;
```
raboczi, line 130: This generates a URL which is sure to contain spaces because of the format of `body`.  Shouldn't it be URL encoded?

raboczi (commented):
My javascript proficiency isn't strong enough that I'm willing to pass or reject this, but I commented on a possible encoding issue.

Apromore-Custom-Plugins/Account-Portal-Plugin/src/main/resources/META-INF/spring/context.xml
```diff
@@ -44,6 +44,9 @@
     <beans:bean id="portalPlugin2" class="org.apromore.plugin.portal.account.DebugPlugin"/>
     <osgi:service ref="portalPlugin2" interface="org.apromore.plugin.portal.PortalPlugin" auto-export="interfaces" />
     -->
+    <osgi:service interface="org.apromore.plugin.portal.PortalPlugin" auto-export="interfaces">
```
raboczi: You can remove the `auto-export` attribute since you're explicitly defining the exported interface here.

raboczi (approved):
I have some suggestions for minor changes, but this appears good enough to merge.

Apromore-Custom-Plugins/Account-Portal-Plugin/src/main/java/org/apromore/plugin/portal/account/ReportIssuePlugin.java
```diff
@@ -0,0 +1,64 @@
...
+    }
+
+    @Override
+    public String getIconPath() {
+        return "/report-issue.svg";
+    }
+
+    @Override
+    public void execute(PortalContext portalContext) {
+        String email = "support@apromore.com";
```
raboczi: This seems like it ought to be configurable rather than a hardcoded value.  I appreciate that configuration is a real nuisance at the moment though, so maybe we can do that later.

PR #845, opened 2021-03-30
AP-3488 Disable timezone and custom calendar

dreamind:


raboczi (approved):
A tidy and proportionate application of force.

PR #848, opened 2021-03-31
AP-3504 Fix share in external plugin

dreamind:


dreamind:
Merge fix in another PR

PR #852, opened 2021-03-31
AP-1924: Updates to service both bpmn.io-based and cytoscape-based log animations

brucenguyen1:
This PR makes adjustments to Apromore-Frontend code to run bpmn.io log animation
- webpack.config.js: generates bpmn.io-based log animation bundle from LogAnimationBpmn class
- Add isBPMNEditor() method to both two wrapper classes: BPMNModelWrapper (for BPMN.io) and GraphModelWrapper (for cytoscape). This is because they have different coordinate systems: bpmn.io has transformation matrix for rotation and other operations while cytoscape only allows zooming and panning. This API method can be improved in the future for a common transparent API to deal with different libraries, but it is enough like this for only bpmn.io and cytoscape.
- Update to TokenAnimation and LogAnimation to use the above changes.

This PR should be tested _after_ this PR: https://github.com/apromore/ApromoreEE/pull/280 has been merged on EE.

For testing: run "npm run build" to generate the bundle file, then deploy dist/loganimationbpmn.js to EE/Log-Animation-Portal-Plugin (see Readme.txt file). Then rebuild and deploy Log-Animation-Portal-Plugin, make sure no changes or errors on UI.


Apromore-Frontend/src/loganimation/tokenAnimation.js
```diff
@@ -66,8 +66,11 @@ export default class TokenAnimation {
         this._TOKEN_LOG_GAP = 3;
```
dreamind, line 66: Consider making some constants static of the class, e.g. TokenAnimation.TOKEN_LOG_GAP

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/static

dreamind (approved):

PR #855, opened 2021-04-01
AP-3512 Add BIMP simuliation server URL to site.properties

franklma:
This PR has a corresponding PR in ApromoreEE
https://github.com/apromore/ApromoreEE/pull/282


raboczi (approved):
May want to consider including the other properties from BIMP portal's client.properties and removing client.properties from BIMP portal.

PR #863, opened 2021-04-06
Feature/AP-3118-add-data-pipelines-to-action-bar

Janite:
Added an icon in the action bar which opens the ETL plugin.

image


Apromore-Core-Components/Apromore-Portal/src/main/java/org/apromore/portal/dialogController/BaseListboxController.java
```diff
@@ -63,20 +63,15 @@
...
 import org.zkoss.zk.ui.event.KeyEvent;
 import org.zkoss.zk.ui.util.Clients;
-import org.zkoss.zul.Button;
-import org.zkoss.zul.ListModelList;
-import org.zkoss.zul.Listbox;
-import org.zkoss.zul.Listhead;
-import org.zkoss.zul.ListitemRenderer;
-import org.zkoss.zul.Messagebox;
-import org.zkoss.zul.Paging;
+import org.zkoss.zul.*;
```
dreamind: I think there is a contentious debate whether to import with wildcard or not. I think if we can adopt one approach consistently, that would be good. I would follow what @raboczi decides on this.

dreamind (approved):

Apromore-Core-Components/Apromore-Portal/src/main/java/org/apromore/portal/dialogController/BaseListboxController.java
raboczi: Checkstyle defaults to "no wildcards" so I'm inclined to go with that.

Janite: Sounds good, will remove the wildcard import in that case.

Apromore-Core-Components/Apromore-Portal/src/main/webapp/index.zul
```diff
@@ -203,6 +203,10 @@
                             <button id="btnUpload" sclass="ap-icon ap-icon-upload" ca:data-t="upload" tooltiptext="${labels.e.portal.upload.hint}"/>
                             <button id="btnDownload" sclass="ap-icon ap-icon-download" ca:data-t="download" tooltiptext="${labels.e.portal.download.hint}"/>
                             <h:span class="ap-action-sep"></h:span>
```
raboczi, line 205: I'd be inclined to include the ETL as part of the same section that includes upload and download and not have the separator between them.  I don't feel strongly about it however, so I'll leave that potential change to your discretion.

raboczi (approved):
This is about as pluggable as could practically be done without significant refactoring.  👌

PR #867, opened 2021-04-07
Made apromoresecuritylib sub-module in ApromoreCore, as per discussio…

fsmicdev:
…n with Simon.


raboczi (approved):
I can confirm the module builds, so since this is existing unedited code I'm not going to read it exhaustively.

PR #873, opened 2021-04-07
AP-3516: Remove "CSV export" from File menu

raboczi:
Removes the CSV export portal plugin from the assembly.  This functionality is already available via the Download menu item.


franklma (approved):
I can confirm "CSV exporter" is removed from the File menu.

PR #881, opened 2021-04-08
Truncate group_usermetadata table to align with restricted viewer logic

franklma:


raboczi (approved):
I'm not the best at interpreting chainfiles, but this seems straightforward enough.

PR #882, opened 2021-04-08
Feature/kc auth int

fsmicdev:


Apromore-Core-Components/Apromore-Portal/src/main/java/org/apromore/portal/security/KeycloakLoginUrlAuthenticationEntryPoint.java
```diff
@@ -104,39 +103,46 @@ protected String buildRedirectUrlToLoginPage(
...
         } else {
-            LOGGER.info("\n\n[ Keycloak SSO turned off ]");
+            LOGGER.info("[ Keycloak SSO turned off ]");
 
             final String requestURI = httpServletRequest.getRequestURI().trim();
-            LOGGER.info("\n\nrequestURI: {}", requestURI);
+            LOGGER.info("requestURI: {}", requestURI);
 
-            if (requestURI.endsWith("/") || requestURI.endsWith("81")) {
+            if (requestURI.endsWith("/") || requestURI.endsWith("81") || requestURI.endsWith("80")) {
```
raboczi: This seems very fragile.  What if the requestURI is "http://www.classof1980", for instance?  Using java.net.URI to parse the requestURI would help.

site.properties
```diff
@@ -17,7 +17,7 @@ site.filestore     = filestore
 site.pql           = pql
 site.logvisualizer = logvisualizer
 
-site.fullProtocolHostPortUrl=http://localhost:8181/
+site.fullProtocolHostPortUrl=http://apromorewebapp.apromoresso.net:8181/
```
raboczi, line 20: Presumably revealing this DNS name isn't a huge issue since it's not a publicly resolvable one, but it does scent slightly of leaking internal security details.  Not opposed to storing these details in site.properties, just maybe not in the default version committed to the open source GitHub repository.

Apromore-Core-Components/Apromore-Portal/src/main/java/org/apromore/portal/security/KeycloakLoginUrlAuthenticationEntryPoint.java
```diff
@@ -67,18 +66,18 @@ public void setKeycloakLoginFormUrl(final String keycloakLoginFormUrl) {
         if ((this.keycloakLoginFormUrl == null) ||
                 (this.keycloakLoginFormUrl.contains(KEYCLOAK_REALM_PLACEHOLDER))) {
             final String keycloakRealm = System.getenv(ENV_KEYCLOAK_REALM_NAME_KEY);
```
raboczi, line 72: The configuration which is currently in environment variables should be moved to site.properties like everything else in the system.  We're already storing the database passwords there, so it's okay (or at least consistent) to store the Keycloak credentials there as well.  Just add a setter method, the property in site.properties, and the relevant beans:property elements to portalContext-security.xml.

site.properties
fsmicdev: Yep, it's actually important because otherwise one docker container can't find the other - but, we can fine tune later I think.

Apromore-Core-Components/Apromore-Portal/src/main/java/org/apromore/portal/security/KeycloakLoginUrlAuthenticationEntryPoint.java
fsmicdev: As discussed, we'll do this sometime near future as Bassi is head under with stuff atm, and we want to get this across the line.

Apromore-Core-Components/Apromore-Portal/src/main/java/org/apromore/portal/security/KeycloakLoginUrlAuthenticationEntryPoint.java
fsmicdev: Rectified with Uri parsing.

raboczi (approved):
We can defer moving the config from environment to site.properties until after Bassi has some time to deal with the changes.  Good enough for now!

PR #883, opened 2021-04-09
Changes for more dynamic/properties based URLs, for things such as th…

fsmicdev:
…e securityms.


raboczi (approved):
God help me, I swore I'd never sat this, but: Looks Good To Me.

PR #884, opened 2021-04-11
AP-3525 Hide obsolete editor menu

dreamind:


brucenguyen1 (approved):
Built with success, tested and see no blue strip occurring during the editor loading, I think these CSS stylesheet changes have made it hidden as expected.

PR #886, opened 2021-04-11
AP-3217

chiichang:
Remove the new criterion if it leads to empty log.
The "criteria" is pointing to currentFilterCriteria at line#98. Hence, there is no need to point currentFilterCriteria = criteria at line#111

### DIFF
--- a/Apromore-Custom-Plugins/Process-Discoverer-Portal-Plugin/src/main/java/org/apromore/plugin/portal/processdiscoverer/impl/apmlog/LogDataWithAPMLog.java
+++ b/Apromore-Custom-Plugins/Process-Discoverer-Portal-Plugin/src/main/java/org/apromore/plugin/portal/processdiscoverer/impl/apmlog/LogDataWithAPMLog.java
@@ -104,11 +104,11 @@ public class LogDataWithAPMLog extends LogData {
     public boolean filter(List<LogFilterRule> criteria) throws Exception {
         this.apmLogFilter.filter(criteria);
         if (apmLogFilter.getPLog().getPTraceList().isEmpty()) { // Restore to the last state
+            criteria.remove(criteria.get(criteria.size() - 1));
             apmLogFilter.filter((List<LogFilterRule>)currentFilterCriteria);
             return false;
         } else {
             this.updateLog(apmLogFilter.getPLog(), apmLogFilter.getApmLog());
-            currentFilterCriteria = criteria;
             return true;
         }
     }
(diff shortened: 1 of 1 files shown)


Apromore-Custom-Plugins/Process-Discoverer-Portal-Plugin/src/main/java/org/apromore/plugin/portal/processdiscoverer/impl/apmlog/LogDataWithAPMLog.java
```diff
@@ -104,11 +104,11 @@ public boolean filterAdditive(LogFilterRule filterCriterion) throws Exception {
...
         if (apmLogFilter.getPLog().getPTraceList().isEmpty()) { // Restore to the last state
+            criteria.remove(criteria.get(criteria.size() - 1));
             apmLogFilter.filter((List<LogFilterRule>)currentFilterCriteria);
             return false;
         } else {
             this.updateLog(apmLogFilter.getPLog(), apmLogFilter.getApmLog());
-            currentFilterCriteria = criteria;
             return true;
         }
     }
```
brucenguyen1: The change above didn't consider LogDataWithAPMLog.clearFilter() which also calls to the filter() method, thus calling clearFilter() will not update currentFilterCriteria leading to error later.

Suggest: please bring back line 111, do not remove it. The reason is because the filter() method is used by other methods, not only filterAdditive(). Inside filterAdditive(), it is true the criteria and currentFilterCriteria point to the same memory space due to the side effect of object reference, but it should not be the main reason to remove line 111 from the filter() method. I hope this is clear enough.

brucenguyen1 (changes requested):

Apromore-Custom-Plugins/Process-Discoverer-Portal-Plugin/src/main/java/org/apromore/plugin/portal/processdiscoverer/impl/apmlog/LogDataWithAPMLog.java
brucenguyen1: Another better change is adding back "currentFilterCriteria = criteria;" to under line 113 (under the IF block), so currentFilterCriteria is always updated to criteria, _no matter_ any side effect in any calling method outside this filter() method

brucenguyen1 (approved):

PR #887, opened 2021-04-12
AP-3296: Fix a bug in log animation logic when it only removes "start" events before the animation

brucenguyen1:
This PR fixes a bug in the log animation logic. For the logs and model in AP-3296, they have "assess" events, but this assess events are not removed but left in the animation making it unexpected alignment. It should have been removed. The consequence is the "Assess application" task is skipped in the animation (tokens flowing on its edge, because the alignment mistakes it as a self-loop which doesn't exist on the model).

There is only one line changed (line 239), other changes were automatic removal of trailing spaces by the editor.

For testing: open AP-3296, choose the log and model in the card to animate, verify that the "Assess application" task is played not skipped (i.e. tokens flowing thought its center, not on its edge). The reason is this log has "assess" and other types of events, but the logic only remove the start.


dreamind (approved):

raboczi (approved):
Code looks fine, functional test confirmed.

PR #892, opened 2021-04-14
AP-3549 AP-3574 Various improvements in CSV Importer

franklma:
AP-3549 Replace logic to detect different types of separators in Log importer 
1. Replace delimiter detection logic inside Log importer with dedicated utils class in _org.apromore.commons.utils_.
2. Modify unit tests accordingly.

AP-3574 Avoid trimming when process headers.


Apromore-Custom-Plugins/CSVImporter-Logic/src/main/java/org/apromore/service/csvimporter/services/MetaDataServiceCSVImpl.java
```diff
@@ -21,55 +21,51 @@
...
+import org.apromore.commons.utils.Delimiter;
+import org.apromore.service.csvimporter.io.CSVFileReader;
+import org.apromore.service.csvimporter.model.LogMetaData;
 
-import java.io.BufferedReader;
-import java.io.IOException;
-import java.io.InputStream;
-import java.io.InputStreamReader;
-import java.io.Reader;
+import java.io.*;
```
raboczi, line 31: At some time in the future when Checkstyle is enabled, I know that one of its rules is to avoid wildcards in import statements.  So we'll save a bit of time later on by not introducing them now.

raboczi (approved):
There's nothing here that should get in the way of a merge.  However I'd like to suggest a check for tab characters in MetaDataServiceCSVImpl.java and replacing them with spaces if there are; some of the indentation looks suspicious.

franklma:
Yep. I was planning to do it in another PR so it not mess with git diff.

PR #896, opened 2021-04-20
AP-3613 Fix NPE in BPMNEdittorController

dreamind:


raboczi (approved):
This solves the issue during BPMN model creation, and I'm happy to approve it.

My comment is that there is abuse of AssertionError, where what would be more appropriate would be just Exception.  Errors are system inconsistencies, and AssertionErrors are for code invariants.

PR #897, opened 2021-04-20
AP-3613 Fix NPE in BPMNEdittorController

dreamind:


raboczi (approved):
Code looks sensible, build, runs and indeed allows new model creation.

PR #899, opened 2021-04-20
AP-3474: OpenXES defaults to using SLF4J

raboczi:
OpenXES has its own custom logging framework.  It defaults to logging to stdout, which is one of the things that needs to be cleaned up as part of AP-3474: all logging should go through the system logger.

While I _could_ have used OpenXES's API to plug in the SLF4J implementation without changing the library, I would have had to do that in several places since many places in Apromore bypass the EventLogService to use OpenXES directly.  I felt that since we control the source code anyway, it would be easiest to just make these light-touch modifications to the library itself.

### DIFF
--- a/Apromore-Extras/OpenXES/src/main/java/org/deckfour/xes/logging/SLF4JLoggingListener.java
+++ b/Apromore-Extras/OpenXES/src/main/java/org/deckfour/xes/logging/SLF4JLoggingListener.java
@@ -75,10 +75,10 @@ public class SLF4JLoggingListener implements XLoggingListener {
 	@Override
 	public void log(String message, Importance importance) {
 		switch (importance) {
-		case DEBUG:   logger.debug(message);   break;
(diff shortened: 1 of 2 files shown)


Apromore-Extras/OpenXES/src/main/java/org/deckfour/xes/logging/SLF4JLoggingListener.java
```diff
@@ -0,0 +1,85 @@
...
+ * 
+ * @author Simon Raboczi (simon.raboczi@apromore.com)
+ */
+public class SLF4JLoggingListener implements XLoggingListener {
+
+	private Logger logger = LoggerFactory.getLogger("OpenXES");
+
+	@Override
+	public void log(String message, Importance importance) {
+		switch (importance) {
```
nolantellis, line 77: This is super cool

nolantellis (approved):

PR #902, opened 2021-04-21
AP-3524-3539-3583-3601-3607 PD and Log Animation improvements

dreamind:
- Also fix broken simulator drawer


Apromore-Custom-Plugins/Process-Discoverer-Portal-Plugin/src/main/java/org/apromore/plugin/portal/processdiscoverer/PDController.java
```diff
@@ -685,6 +689,7 @@ else if (newMode == InteractiveMode.TRACE_MODE) {
             timeStatsController.setDisabled(true);
             toolbarController.setDisabled(true);
             toolbarController.setDisabledAnimation(true);
+            toolbarController.toogleAnimateBtn(true);
```
brucenguyen1: Call toolbarController.toogleAnimateBtn(true) should not be used here because animation is not allowed when viewing a trace: "Animate log" button is disabled (the method call is right above).

Apromore-Core-Components/Apromore-Portal/src/main/java/org/apromore/portal/dialogController/BaseController.java
```diff
@@ -223,14 +223,18 @@ protected String getExportPath(final String nativeType) {
 
     public void toggleComponentSclass(HtmlBasedComponent comp, boolean state, String stateOff, String stateOn) {
         String sclass = Objects.requireNonNull(comp.getSclass(), stateOff);
```
brucenguyen1: The error message in case comp.getSclass() should be meaningful instead of stateOff, e.g. "The Sclass value of component id=" + comp.getId + " must be not null".

brucenguyen1: Is it better if we just handle null value instead of throwing NullPointerException for this method?
if (comp) == null return;
if (comp.getSclass() == null) sclass = ""

brucenguyen1: In the same vein of input validation above, we can also add handling for stateOn and stateOff because NullPointerException will be thrown in String.contains. NullPointerException will lead to unexpected error message from the system.
if (stateOn == null || stateOff == null) return;

brucenguyen1 (commented):
Built and the features work fine. Just some minor comments to make it a bit safer.

brucenguyen1 (approved):

PR #906, opened 2021-04-22
AP-3296: Wrong message shown due to using a wrong condition

brucenguyen1:
Should check edges of SequenceFlow type only in the model as there are other Edge instances such as Data Association which are not used by the animation.


raboczi (approved):
Since this is private to the log animation, you're entitled to decide what's an animatable model or not.  I would have been inclined to leave the existing check that all edges are connected at at least one end and make the more stringent check that sequence flows are connected at both ends conditionally on that.  Since the log animation ignores all other edges however, this is fine for that purpose.

PR #908, opened 2021-04-22
AP-3574 Avoid trimming when process headers in LogImporterCSVImpl

franklma:
Avoid trimming empty headers since OpenCSV doesn't remove empty columns.


chiichang (approved):
build&run without problem

PR #909, opened 2021-04-23
AP-3582: Automatic replay of the animation after the model has stopped moving or zooming

brucenguyen1:
As cytoscape doesn't support built-in events such as movingStop or zoomingStop (like bpmn.io), a custom solution is created to do the same. The moving stop is detected based on mouseup with a moving state remembered from the panning (i.e. moving) event. The zooming stop is implemented via jquery event extensions for mouse wheel events and the event is checked on top of the graph div container.
Changes:
- Add jquey event extension for mouse wheel start and stop events
- Register query event extension when starting the animation
- Add event handlers for mouseup and wheelstop events when initializing graph model wrapper.

### DIFF
--- a/Apromore-Frontend/src/commons/jqueryExtensions.js
+++ b/Apromore-Frontend/src/commons/jqueryExtensions.js
@@ -1,0 +1,73 @@
+
+// 'Special scroll events for jQuery' by James Padolsey
+// https://j11y.io/javascript/special-scroll-events-for-jquery/
+export const registerWheelStartStopEvents = function(jQuery) {
+    (function () {
+        let special = jQuery.event.special,
+            uid1 = 'D' + (+new Date()),
+            uid2 = 'D' + (+new Date() + 1);
+
+        special.wheelstart = {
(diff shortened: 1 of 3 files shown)


Apromore-Frontend/src/commons/jqueryExtensions.js
```diff
@@ -0,0 +1,73 @@
...
+            setup: function () {
+
+                var timer,
+                    handler = function (evt) {
+
+                        var _self = this,
+                            _args = arguments;
+
+                        if (timer) {
+                            clearTimeout(timer);
```
dreamind, line 51: To be safe, probably add timer = null

Apromore-Frontend/src/commons/jqueryExtensions.js
```diff
@@ -0,0 +1,73 @@
...
```
dreamind, line 20: To be safe, probably add timer = null

dreamind (approved):

PR #912, opened 2021-04-25
AP-3575 AP-3604 Improvements to user deleting

franklma:
AP-3575 Fix error when deleting user and choose "deleting all assets". If user_to_be_deleted owns two nested folders, the child folder needs to be deleted first.

AP-3604 During user deleting, disable "Delete assets" if a folder that solely owned by User-To-Be-Deleted with a file co-owned within.


chiichang (approved):
build&run without problem

PR #925, opened 2021-04-28
AP-1924: move animation package from Portal Plugin module to Logic module.

brucenguyen1:
This PR is to move the "animation" package from Process-Discoverer-Portal-Plugin to Log-Animation-Logic. It contains animation recording logic and should be placed in this logic module. It is commonly used by PD and BPMN.io-based Log-Animation-Portal-Plugin.
Changes:
- Move source code, test code, test data together
- Changes to pom.xml and context.xml 
- Remove references from PD Portal to Log-Animation-Portal-Plugin interface because it plays animation directly now.

To check: build, deploy and run animation in PD and Portal, make sure all works fine.

Note: this PR must use together another PR of the same name in EE.


raboczi (approved):
This looks great.

PR #928, opened 2021-04-29
AP-3242: Changes to make token flow more fluid

brucenguyen1:
Changes to make token flow more fluid: 
- Refine the token clustering technique: adjacent/overlapping tokens are merged based on an absolute distance measure to avoid small number issue (the distance now ranges from 1 to 36,000 based on the number of frames). Before it was based on relative distance (i.e. the percentage distance value from 0 to 1) which makes it inaccurate to define the closeness measure between adjacent tokens, e.g. with closeness measure = 0.01, tokens with distance = 0.001 and 0.005 (diff = 0.004) will be always considered as being close, while tokens with distance = 0.8 and 0.9 are not considered as being close.
- Refine the radius calculation for the token circle, now is based on logarithm function compared to a step-wise function before.

### DIFF
--- a/Apromore-Custom-Plugins/Log-Animation-Logic/src/main/java/org/apromore/service/loganimation/recording/Frame.java
+++ b/Apromore-Custom-Plugins/Log-Animation-Logic/src/main/java/org/apromore/service/loganimation/recording/Frame.java
@@ -128,8 +128,8 @@ public class Frame {
     private double getFrameIndexRelativeTokenDistance(int logIndex, int tokenIndex) {
         int startFrameIndex = animationIndexes.get(logIndex).getStartFrameIndex(tokenIndex);
         int endFrameIndex = animationIndexes.get(logIndex).getEndFrameIndex(tokenIndex);
-        int numberOfFrames = endFrameIndex - startFrameIndex;
-        return (numberOfFrames == 0) ? 0 : (double)(frameIndex - startFrameIndex)/numberOfFrames;
+        int maxLength = endFrameIndex - startFrameIndex;
+        return (maxLength == 0) ? 0 : (double)(frameIndex - startFrameIndex)/maxLength;
     }
     
     /**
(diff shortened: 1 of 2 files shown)


Apromore-Custom-Plugins/Log-Animation-Logic/src/main/java/org/apromore/service/loganimation/recording/Frame.java
```diff
@@ -121,12 +121,26 @@ public void removeToken(int logIndex, int tokenIndex) {
...
-    public double getTokenDistance(int logIndex, int tokenIndex) {
+    /**
+     * This is the percentage from the start of the element (0..1) based on frame indexes
+     * This distance is suitable for the relative position of tokens on a modeling element
+     */
+    private double getFrameIndexRelativeTokenDistance(int logIndex, int tokenIndex) {
         int startFrameIndex = animationIndexes.get(logIndex).getStartFrameIndex(tokenIndex);
-        int numberOfFrames = animationIndexes.get(logIndex).getEndFrameIndex(tokenIndex) - startFrameIndex;
+        int endFrameIndex = animationIndexes.get(logIndex).getEndFrameIndex(tokenIndex);
+        int numberOfFrames = endFrameIndex - startFrameIndex;
```
raboczi: Is there an off-by-one error here?  If there is one frame at index 0 which is both the start frame and the end frame, then subtracting the two indices will give zero, not one.

raboczi (approved):
Builds and runs, looks tidy, approved.  One comment about a possible off-by-one error that's more likely just me being misled by variable names.

Apromore-Custom-Plugins/Log-Animation-Logic/src/main/java/org/apromore/service/loganimation/recording/Frame.java
brucenguyen1: Ok. The variable name "numberOfFrames" is not totally accurate, changed it to maxLength. The purpose is when frameIndex = startFrameIndex, the return value is 0, when frameIndex = endFrameIndex, the return value is 1. The return value ranges from 0 to 1.

PR #930, opened 2021-04-30
AP-3578: Make SplitMiner not to produce mixed split-join gateways

brucenguyen1:
This PR is to update SplitMiner not to produce mixed split-join gateways, i.e. those with multiple incoming arcs and multiple outgoing arcs. It is because the semantics of these gateways is usually ill-defined and not suitable for log animation.

The commented code remains because this is a complex algorithm from PhD research, need to keep track of the changes to this code.

Verify: open PurchasingExample in PD, push the arc slider all the way to the right hand side, view BPMN diagram. Make sure the first and second gateways are not merged, they should be one join-XOR and one split-XOR separately.

### DIFF
--- a/Apromore-Custom-Plugins/SplitMiner-Logic/src/main/java/org/apromore/splitminer/SplitMiner.java
+++ b/Apromore-Custom-Plugins/SplitMiner-Logic/src/main/java/org/apromore/splitminer/SplitMiner.java
@@ -187,12 +187,12 @@ public class SplitMiner {
             e.printStackTrace();
             try{
             	dfgp.resetDFGPStructures();
-                dfgp.buildSafeDFGP(); //recreate a safe data structure for the graph in case errors happen (the graph could be disconnected)  
+                dfgp.buildSafeDFGP(); //recreate a safe data structure for the graph in case errors happen (the graph could be disconnected)
                 transformDFGPintoBPMN();
                 if (structuringTime == SplitMinerUIResult.StructuringTime.POST) structure();
             } catch ( Exception ee ) {
                 System.out.println("ERROR - something went wrong in translating DFG to BPMN");
-                e.printStackTrace();            
+                e.printStackTrace();
                 throw e;
             }
         }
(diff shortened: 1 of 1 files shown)


Apromore-Custom-Plugins/SplitMiner-Logic/src/main/java/org/apromore/splitminer/SplitMiner.java
```diff
@@ -187,12 +187,12 @@ public BPMNDiagram discoverFromDFGP(DirectlyFollowGraphPlus idfgp) throws Except
...
             try{
             	dfgp.resetDFGPStructures();
-                dfgp.buildSafeDFGP(); //recreate a safe data structure for the graph in case errors happen (the graph could be disconnected)  
+                dfgp.buildSafeDFGP(); //recreate a safe data structure for the graph in case errors happen (the graph could be disconnected)
                 transformDFGPintoBPMN();
                 if (structuringTime == SplitMinerUIResult.StructuringTime.POST) structure();
             } catch ( Exception ee ) {
                 System.out.println("ERROR - something went wrong in translating DFG to BPMN");
-                e.printStackTrace();            
+                e.printStackTrace();
```
raboczi, line 195: Nothing you changed, but this is logging to stdout rather than using the system logger.

Apromore-Custom-Plugins/SplitMiner-Logic/src/main/java/org/apromore/splitminer/SplitMiner.java
```diff
@@ -346,13 +346,16 @@ private void transformDFGPintoBPMN() {
...
+        // Bruce 30.04.2021: comment the below to produce diagrams without mixed split-join gateways
+        /*
         if( replaceIORs ) {
             helper.collapseSplitGateways(bpmnDiagram);
             helper.collapseJoinGateways(bpmnDiagram);
         } else {
             helper.collapseSplitGateways(bpmnDiagram);
             helper.collapseJoinGateways(bpmnDiagram);
         }
+        */
```
raboczi, line 358: Glancing over the rest of the code, you seem to have put some effort into making `replaceIORs` a configurable option.  Do you have a reason not to do it as
```
if( replaceIORs ) {
    helper.collapseSplitGateways(bpmnDiagram);
    helper.collapseJoinGateways(bpmnDiagram);
}
```
and pass `replaceIORs` as false?

raboczi (approved):

PR #935, opened 2021-05-04
AP-3682: Sanitize BPMN uploads

raboczi:
This is PR modifies the ProcessService within the Manager to sanitize uploaded BPMN files.  The very specific transformation which is performed is that any element appearing as the child of a bpmn:text element is converted into text, e.g.
```
<bpmn:text><script>alert("Boo")</script></bpmn:text>
```
will be rewritten as
```
<bpmn:text>&lt;script&gt;alert("Boo")&lt;/script&gt;</bpmn:text>
```
in order to prevent script injection attacks.  The actual work is done by sanitizeBPMN.xsl.

Sanitization can be enabled or disabled from site.properties using the `manager.sanitization.enable` flag.  There are some rather brittle unit tests, but to do better would require more work writing test utilities for comparing XML streams.

### DIFF
--- a/Apromore-Core-Components/Apromore-Manager/src/test/java/org/apromore/test/service/impl/ProcessServiceImplUnitTest.java
+++ b/Apromore-Core-Components/Apromore-Manager/src/test/java/org/apromore/test/service/impl/ProcessServiceImplUnitTest.java
@@ -919,6 +919,32 @@ public class ProcessServiceImplUnitTest extends EasyMockSupport {
   }
 
 
+  /**
+   * Test the {@link ProcessServiceImpl#sanitizeBPMN} method.
+   */
+  @Test
+  public void testSanitizeBPMN() throws Exception {
+
+      // Check that nothing changes for an innocuous BPMN file
+      Assert.assertEquals(
+          CharStreams.toString(new InputStreamReader(getResourceAsStream("BPMN_models/Cyclic.bpmn"))),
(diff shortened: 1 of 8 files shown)


Apromore-Core-Components/Apromore-Manager/src/test/java/org/apromore/test/service/impl/ProcessServiceImplUnitTest.java
```diff
@@ -919,6 +919,32 @@ public void testDeleteProcessModel_BranchHasOnlyOnePMV() throws Exception {
...
+  /**
+   * Test the {@link ProcessServiceImpl#sanitizeBPMN} method.
+   */
+  @Test
+  public void testSanitizeBPMN() throws Exception {
+
+      // Check that nothing changes for an innocuous BPMN file
+      Assert.assertEquals(
+          CharStreams.toString(new InputStreamReader(getResourceAsStream("BPMN_models/Cyclic.bpmn"))),
+          CharStreams.toString(new InputStreamReader(ProcessServiceImpl.sanitizeBPMN(getResourceAsStream("BPMN_models/Cyclic.bpmn")))) + "\n"
```
raboczi, line 931: The XSL transformation elides the trailing newline, which is why I have to manually append it here.  It'd be helpful for someone on Windows to confirm the XML processing isn't sensitive to the system line ending.

nolantellis (approved):

PR #936, opened 2021-05-04
AP-402 Detail tooltip in case detail view

dreamind:


Apromore-Custom-Plugins/Process-Discoverer-Portal-Plugin/src/main/java/org/apromore/plugin/portal/processdiscoverer/components/CaseDetailsController.java
```diff
@@ -78,6 +87,37 @@ public String getExportFilename() {
 		return parent.getContextData().getLogName() + ".csv";
 	}
 
+	/**
+	 * Update Activity <-> Attributes Map for quick lookup
+	 **/
+	private void updateActivityToAttributeMap(String caseId) {
+		activityToAttributeMap.clear();
```
nolantellis, line 93: Can we move this method into some service class that already exists for eventlogs and usr tht service in this controller. 
We should avoid business logic here.

Apromore-Custom-Plugins/Process-Discoverer-Portal-Plugin/src/main/java/org/apromore/plugin/portal/processdiscoverer/components/CaseDetailsController.java
```diff
@@ -78,6 +87,37 @@ public String getExportFilename() {
...
+		APMLog log = ((LogDataWithAPMLog) parent.getLogData()).getFilteredAPMLog();
+		ATrace trace = log.get(caseId);
+		List<AActivity> activities = trace.getActivityList();
+		for (int i = 0; i < activities.size(); i++) {
+			AActivity activity = activities.get(i);
+			String activityName = activity.getName();
```
nolantellis: move this to a service class. and return a JSONObject . and use toString oin this controller at the end.

nolantellis (changes requested):
Need some minor changes

Apromore-Custom-Plugins/Process-Discoverer-Portal-Plugin/src/main/java/org/apromore/plugin/portal/processdiscoverer/components/CaseDetailsController.java
dreamind: Actually, there is no service associated with this component. And the map is UI construct, so it's more like a data transform into ViewModel rather than Model itself.

nolantellis: but its logic again. I think the UI should just use the objects given from service. Example the apmLog is coming from some service class. Can we do the logic there and ust use the object here.

brucenguyen1: Agree. This is the interaction on UI only. The interaction is like this: when user views the list of Cases, a window popup will show this list of cases in a table, each row is one case. When the user clicks on each row (a case), a linear visualization of the case is displayed in the form of a graph in PD (i.e. a straight line of boxes, each box is one event). Next, when the user hovers the mouse over one box (one event), there is a list of standard attributes of this event is shown in a tooltip box. The current implementation is using AJAX and JSON to send the updated event attributes for each case whenever the user clicks on a row in the table, it uses ZK's AJAX engine to do that.

Apromore-Custom-Plugins/Process-Discoverer-Portal-Plugin/src/main/java/org/apromore/plugin/portal/processdiscoverer/components/CaseDetailsController.java
brucenguyen1: Because the trace visualization is done by ALog and ATrace, it'll be more consistent and safe to get this event attributes from ALog/ATrace rather than APMLog.

brucenguyen1 (approved):

nolantellis (approved):

PR #955, opened 2021-05-10
AP-3759 Hide change password for keycloak

dreamind:


Apromore-Core-Components/Apromore-Portal/src/main/java/org/apromore/portal/dialogController/UserMenuController.java
```diff
@@ -120,7 +121,11 @@ public void doAfterCompose(Menubar menubar) {
                 }
 
                 String menuName = plugin.getGroupLabel(Locale.getDefault());
+                String label = plugin.getLabel(Locale.getDefault());
 
+                if (label == "Change password" && usingKeycloak) {
+                    continue;
+                }
```
raboczi: The ChangePasswordPlugin class should be overriding the getAvailability method to return PortalPlugin.Availability.UNAVAILABLE to do this, so that the policy resides in the plugin rather than being coded as an ad hoc check in the menu provider.

raboczi (changes requested):
Unless there's some reason it can't be done that way, I'd prefer this check to be part of the ChangePasswordPlugin instead.

Apromore-Core-Components/Apromore-Portal/src/main/java/org/apromore/portal/dialogController/UserMenuController.java
raboczi: Here's my concrete suggestion for the method to add to ChangePasswordPlugin:
```
import org.apromore.plugin.portal.PortalPlugin.Availability;
import org.apromore.portal.ConfigBean;
import org.zkoss.spring.SpringUtil;

// ...

    @Override
    public Availability getAvailability() {
        ConfigBean config = (ConfigBean) SpringUtil.getBean("portalConfig");

        return config.isUseKeycloakSso() ? Availability.UNAVAILABLE : Availability.AVAILABLE;
    }
```

raboczi (approved):

PR #957, opened 2021-05-11
 AP-3682: More sanitization tweaks

raboczi:
Previously, the sanitizer script would fail if the suspicious XML element (e.g. `<script>`) had any attributes, since it would try to add the attribute to the text (e.g. `&lt;script&gt;`) and text can't have attributes.  Attributes are now correctly converted to text.

This PR has no other dependencies and doesn't break the existing tests.

### DIFF
--- a/Apromore-Core-Components/Apromore-Manager/src/main/resources/xsd/sanitizeBPMN.xsl
+++ b/Apromore-Core-Components/Apromore-Manager/src/main/resources/xsd/sanitizeBPMN.xsl
@@ -8,16 +8,15 @@
 <!-- Disallow any complex content (e.g. <img>, <script>) in human-legible elements by entering textOnly mode. -->
 <xsl:template match="bpmn:documentation|bpmn:text">
     <xsl:copy>
-        <xsl:apply-templates select="@*|node()" mode="textOnly"/>
+        <xsl:apply-templates select="@*"/>
+        <xsl:apply-templates select="node()" mode="textOnly"/>
     </xsl:copy>
 </xsl:template>
 
 <!-- Any <script> tags outside of textOnly mode will still be executed by bpmn.io, so we comment them out. -->
 <xsl:template match="*[local-name()='script']">
     <xsl:comment>
-        <xsl:text>&lt;script&gt;</xsl:text>
-        <xsl:apply-templates select="@*|node()" mode="textOnly"/>
-        <xsl:text>&lt;/script&gt;</xsl:text>
+        <xsl:apply-templates select="." mode="textOnly"/>
     </xsl:comment>
 </xsl:template>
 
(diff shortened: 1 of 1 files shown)


Apromore-Core-Components/Apromore-Manager/src/main/resources/xsd/sanitizeBPMN.xsl
```diff
@@ -8,16 +8,15 @@
 <!-- Disallow any complex content (e.g. <img>, <script>) in human-legible elements by entering textOnly mode. -->
 <xsl:template match="bpmn:documentation|bpmn:text">
     <xsl:copy>
-        <xsl:apply-templates select="@*|node()" mode="textOnly"/>
+        <xsl:apply-templates select="@*"/>
+        <xsl:apply-templates select="node()" mode="textOnly"/>
```
raboczi, line 12: This change means that attributes of the bpmn:text or bpmn:documentation element will still be present; only the content inside the element needs to be sanitized.

Apromore-Core-Components/Apromore-Manager/src/main/resources/xsd/sanitizeBPMN.xsl
```diff
@@ -8,16 +8,15 @@
...
     </xsl:copy>
 </xsl:template>
 
 <!-- Any <script> tags outside of textOnly mode will still be executed by bpmn.io, so we comment them out. -->
 <xsl:template match="*[local-name()='script']">
     <xsl:comment>
-        <xsl:text>&lt;script&gt;</xsl:text>
-        <xsl:apply-templates select="@*|node()" mode="textOnly"/>
-        <xsl:text>&lt;/script&gt;</xsl:text>
+        <xsl:apply-templates select="." mode="textOnly"/>
```
raboczi, line 19: Avoid duplicating the rule for rendering an element as text now that it's a bit more complicated; we just reparse the script element using the textOnly template.

Apromore-Core-Components/Apromore-Manager/src/main/resources/xsd/sanitizeBPMN.xsl
```diff
@@ -31,17 +30,19 @@
...
 
-<!-- Inside an element which is disallowed complex content, attributes and text get passed through unchanged. -->
-<xsl:template match="@*|text()" mode="textOnly">
-    <xsl:copy>
-        <xsl:apply-templates select="@*|node()" mode="textOnly"/>
-    </xsl:copy>
+<!-- Inside an element which is disallowed complex content, text gets passed through unchanged. -->
+<xsl:template match="text()" mode="textOnly">
+    <xsl:value-of select="."/>
 </xsl:template>
```
raboczi, line 36: Since we know that XML text nodes are leaf nodes, we can do without the recursive machinery.

nolantellis (approved):

PR #967, opened 2021-05-12
AP-2860 Put extra check for user management dirty flag

dreamind:


Apromore-Custom-Plugins/User-Admin-Portal-Plugin/src/main/java/org/apromore/plugin/portal/useradmin/UserAdminController.java
```diff
@@ -777,6 +777,7 @@ private void setSelectedUsers(Set<User> users) {
         }
         updateAssignedRoleModel(users);
         updateAssignedGroupModel(users);
+        isUserDetailDirty = false; // extra defensive, ensure dirty is not set by field's setValue
```
raboczi: Would it make sense to remove the duplicate `isUserDetailDirty = false;` on line 730?  It just seems odd to set it twice in the same method.  The convention seems to be to set the dirty flags at the end of methods that mutate or update the current user or group.

Apromore-Custom-Plugins/User-Admin-Portal-Plugin/src/main/java/org/apromore/plugin/portal/useradmin/UserAdminController.java
```diff
@@ -808,6 +809,7 @@ private Group setSelectedGroup(Group group) {
         nonAssignedUserModel.setMultiple(true);
         nonAssignedUserList = new AssignedUserListbox(nonAssignedUserListbox, nonAssignedUserModel, "Users not in the group");
         selectedGroup = group;
+        isGroupDetailDirty = false; // extra defensive, ensure dirty is not set by field's setValue
```
raboczi: Similarly, would it make sense to remove the duplicate `isGroupDetailDirty = false;` on line 784?

Apromore-Custom-Plugins/User-Admin-Portal-Plugin/src/main/java/org/apromore/plugin/portal/useradmin/UserAdminController.java
```diff
@@ -265,7 +265,7 @@ public void doAfterCompose(Window win) throws Exception {
         mainWindow = win;
         String userName = portalContext.getCurrentUser().getUsername();
         currentUser = securityService.getUserByName(userName);
-        selectedUser = currentUser;
+        selectedUser = null;
```
raboczi, line 268: While this change isn't harmful, it seems a little less convenient that the old default of showing someone their own account when they open the user management window.

raboczi (approved):
I don't see anything harmful to warrant not approving this merge, but I have a few comments/queries.

Apromore-Custom-Plugins/User-Admin-Portal-Plugin/src/main/java/org/apromore/plugin/portal/useradmin/UserAdminController.java
dreamind: Actually somewhere after that, we have setSelectedUsers(null); I think for now I just want to make sure nothing is selected to limit potential causes if the issue remains.

PR #968, opened 2021-05-13
AP-3641 Bug fix patch

franklma:


raboczi (approved):
This makes sense to me.

PR #983, opened 2021-05-17
AP-3474: Better logging

raboczi:
This PR overhauls logging throughout the system.
* There is a new factory class for loggers called PortalLoggerFactory.  Loggers it creates always set an apromore.user attribute which then shows up as new column in the system log file.  Note that the new column doesn't appear when using the Karaf console log:* commands; you have to use the log file in assembly/data/log/karaf.log directly.
* In the ZK presentation layer (i.e. the Portal and its plugins) the stock SLF4J LoggerFactory is swapped out for PortalFactory.  This is why there are so many files in this PR.  For the bulk of these, the only change is the replacement of SLF4J LoggerFactory with PortalLoggerFactory in the imports list and LOGGER variable declaration.
* New logging messages have been judiciously added so that most common user operations record who did which action to which process model, log, or folder.

To functionally test this, watch the karaf.log (e.g. using `tail -f enterprise-assemblies/apromore-ee/target/assemby/data/log/karaf.log`) while using the system.  The second-last column of the raw log should record who the user is during login, editing, etc.  This only works reliably in the presentation layer because the business logic layer and below doesn't have access to the user session.

### DIFF
--- a/Apromore-Core-Components/Apromore-Portal/src/main/java/org/apromore/portal/dialogController/workspaceOptions/AddFolderController.java
+++ b/Apromore-Core-Components/Apromore-Portal/src/main/java/org/apromore/portal/dialogController/workspaceOptions/AddFolderController.java
@@ -25,8 +25,8 @@
 package org.apromore.portal.dialogController.workspaceOptions;
 
 import java.io.IOException;
-import java.util.logging.Logger;
 
+import org.slf4j.Logger;
 import org.zkoss.zk.ui.Execution;
 import org.zkoss.zk.ui.Executions;
 import org.zkoss.zk.ui.event.Event;
@@ -38,6 +38,7 @@ import org.zkoss.zul.*;
 
 import org.apromore.dao.model.User;
 import org.apromore.exception.NotAuthorizedException;
+import org.apromore.plugin.portal.PortalLoggerFactory;
 import org.apromore.portal.common.ItemHelpers;
 import org.apromore.portal.common.UserSessionManager;
 import org.apromore.portal.common.notification.Notification;
@@ -53,7 +54,7 @@ public class AddFolderController extends BaseController {
     private Button btnSave;
     private Button btnCancel;
     private Textbox txtName;
-    private Logger LOGGER = Logger.getLogger(AddFolderController.class.getCanonicalName());
+    private static final Logger LOGGER = PortalLoggerFactory.getLogger(AddFolderController.class);
 
     public AddFolderController(MainController mainController, User currentUser, FolderType currentFolder) throws DialogException {
         this.mainController = mainController;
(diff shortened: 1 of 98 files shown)


Apromore-Core-Components/Apromore-Portal/src/main/java/org/apromore/portal/dialogController/workspaceOptions/AddFolderController.java
```diff
@@ -118,24 +119,22 @@ private void save() throws InterruptedException {
...
-            LOGGER.warning("folderName " + folderName);
+            LOGGER.info("Creating folder " + folderName);
             String userId = UserSessionManager.getCurrentUser().getId();
             FolderType currentFolder = this.mainController.getPortalSession().getCurrentFolder();
             int currentParentFolderId = currentFolder == null || currentFolder.getId() == 0 ? 0 : currentFolder.getId();
             this.mainController.getService().createFolder(userId, folderName, currentParentFolderId);
             this.mainController.reloadSummaries();
+
+        } catch (WrongValueException ex) {
+            // Messagebox.show("You have entered invalid value.", "Apromore", Messagebox.OK, Messagebox.ERROR);
```
Janite, line 131: Should this exception also be logged if caught?

Apromore-Core-Components/Apromore-Portal/src/main/java/org/apromore/portal/dialogController/workspaceOptions/RenameFolderController.java
```diff
@@ -104,23 +105,20 @@ private void save() throws InterruptedException {
...
             }
 
-            LOGGER.warning("folderName " + folderName);
+            LOGGER.info("Rename folder " + folderName);
             this.mainController.getService().updateFolder(this.folderId, folderName, UserSessionManager.getCurrentUser().getUsername());
             this.mainController.reloadSummaries();
             this.folderEditWindow.detach();
+
+        } catch (WrongValueException ex) {
+            // Messagebox.show("You have entered invalid value.", "Apromore", Messagebox.OK, Messagebox.ERROR);
```
Janite, line 115: Same as above.

Apromore-Core-Components/Apromore-Portal/src/main/java/org/apromore/portal/dialogController/workspaceOptions/AddFolderController.java
raboczi: In this case I was just trying to preserve the original behavior, but that doesn't mean the original behavior was necessarily correct.  Could even be useful for someone interested in usability to log the inputs that users get wrong.  I'll log it at debug level.

Apromore-Core-Components/Apromore-Portal/src/main/java/org/apromore/portal/dialogController/workspaceOptions/RenameFolderController.java
raboczi: Also logged at debug level.

Janite (approved):

PR #986, opened 2021-05-18
Apromore-core won't start without the JWT libraries.

raboczi:
This PR fixes a reversion that prevented ApromoreCore from being started on its own.  It adds the now-required JWT support bundle.  Test by executing ApromoreCore/core-assemblies/apromore-core/target/assembly/bin/karaf and confirming that the server can be started.


franklma (approved):
Compile and run Core successfully.

image

PR #1008, opened 2021-05-26
AP-3867: Synchronizing JWT security token management with ZK activity

raboczi:
This PR corrects the management of the Keycloak-issued security cookie, a JSON web token (JWT).  When Keycloak authentication is enabled, we issue these tokens to users when they log in, and we require them to present the token (as a browser cookie) during all interactions with the GUI.  Previously, there were two issues:
* The token was only checked during access to the Portal, but not during access to any other tool opening in a different tab.
* The token would expire after 30 minutes, ending the user's session.  User activity did not keep the session alive.

The first point has been addressed by deleting the JWT management logic from MainController (i.e. the index.zul page) and relocating it into a new class, ApromoreZKListener.  This class receives a callback every time ZK receives an AJAX request from the browser, no matter which window/tab the traffic is directed to.

The second point has been addressed by adding a check to see if the user's security token is close to expiring, and if so to replace it with a fresh one, identical except that the expiry date is again 30 minutes from when it's issued.  "Close to expiring" is configured by a new property in site.properties called site.minutesUntilExpiryBeforeSessionRefresh with a default of 20 minutes.  This will result in the user's JWT being recreated every 10 minutes during active use.

A collateral feature is that the Jersey 2 library was added to ApromoreCore to enable ApromoreZKListener to fetch the refreshed security tokens from the security microservice via REST.   In a subsequent PR this will allow the REST endpoints in ApromoreEE to be upgraded from Jersey 1 to Jersey 2, and [the existing Jersey 1 bundle can be removed](https://apromore.atlassian.net/browse/AP-3314).

This PR also fixes a defect Nolan spotted in KeycloakLoginUrlAuthenticationEntryPoint.java.  URLs without a port were treated as port -1 rather than the protocol default (e.g. 80 for HTTP).


Apromore-Core-Components/Apromore-Portal/src/test/java/org/apromore/portal/security/helper/JwtHelperUnitTest.java
```diff
@@ -0,0 +1,90 @@
...
+     * Test the {@link JwtHelper#getClaimsSetFromJWT} method.
+     *
+     * This test shows up a problem that may originate with the security microservice, that
+     * it's issuing JWTs with time-valued fields encoded as milliseconds since the unix
+     * epoch rather than seconds since the unix epoch.
+     */
+    @Test
+    public void testGetClaimsSetFromJWT() throws Exception {
+
+        final String testJWT = "eyJhbGciOiJSUzI1NiJ9.eyJzdWIiOiJhZG1pbiIsImdpdmVuTmFtZSI6IlRlc3QiLCJmYW1pbHlOYW1lIjoiVXNlciIsImV4cCI6MTYxOTE1NDgyNTkwMywiaWF0IjoxNjE5MTQ3NjI1OTAzLCJlbWFpbCI6IiIsImtjdXNlcmlkIjoiZjoyY2JjM2E2NS0yNzNhLTRjMTEtYjFlNC0zZjhkNjVlM2JiYjg6OCJ9.XtEHtIevrp7PVbYLwGbjBbvB32kZ8ZM2au-fbKGn03zjLYXojibRXbLfqJb1p45WbEn9nvoGn-oFvmSecWQsQGBSF3iGhy1RjpPgKuZxMdKBaGDaLdqix0OgQYFOk2RLaSRoBTVEdd579UCIavF66vVW9rKCC4zE2AIFSFFFPgJK25lwq7c31DyYjs5deA_SlmV5z1tfkk27ksrOuVmy-LzmAf81fFp5OTftTvM3Zfb04AmKBnMLyIuSb63zdlRlHrgjVbsE29CIr4u0NIdNt-hOrjXXe_j8PMO26tsycG0ku2Fg8ZYcvozWeTa3uvH4dvYN7Dsrn44A9P2tDwze8w";
```
raboczi: Replace this testJWT with one that has correctly encoded (i.e. seconds since the unix epoch, rather than milliseconds) Date claims.

Apromore-Core-Components/Apromore-Portal/src/test/java/org/apromore/portal/security/helper/JwtHelperUnitTest.java
```diff
@@ -0,0 +1,90 @@
...
+
+public class JwtHelperUnitTest {
+
+    private static final String VALID_KC_ISSUER_URL = "http://keycloak.apromoresso.net";
+    private static final String VALID_TENANT_IP_ADDRESS = "203.20.10.4";
+
+    private static final String SUBJECT_UUID = UUID.randomUUID().toString();
+    private static final String SUBJECT_EMAIL_ADDRESS = "mic.giansiracusa@apromore.com";
+    private static final String FIRST_NAME = "Mic";
+    private static final String SURNAME = "G";
```
raboczi: These constants are no longer used and should be removed.

Apromore-Core-Components/Apromore-Portal/pom.xml
```diff
@@ -24,6 +24,7 @@
 
     <properties>
         <junit.jupiter.version>5.7.0</junit.jupiter.version>
+        <keycloak.version>12.0.4</keycloak.version>
```
raboczi: This property is inherited already from ApromoreCore/pom.xml, so I think it's redundant here.

Apromore-Core-Components/Apromore-Portal/src/main/java/org/apromore/portal/security/helper/JwtHelper.java
```diff
@@ -38,32 +38,47 @@
...
 import java.time.LocalDateTime;
 import java.time.ZoneId;
 import java.util.Date;
+import static org.apromore.portal.util.AssertUtils.notNullAssert;
 
 public class JwtHelper {
 
     public static final String JWT_KEY_SUBJECT_USERNAME = "sub";
+    public static final String JWT_KEY_ISSUED_AT = "iat";
+    public static final String JWT_EXPIRY_TIME = "exp";
```
raboczi, line 57: These constants duplicate JWTClaimsSet.SUBJECT_CLAIM, ISSUED_AT_CLAIM and EXPIRATION_TIME_CLAIM.  See line 84+ of the source code at https://github.com/felx/nimbus-jose-jwt/blob/master/src/main/java/com/nimbusds/jwt/JWTClaimsSet.java

fsmicdev:
They're private scoped in JwtClaimsSet, so can't be accessed (and, hence
not used in key assignment of builder)... however, I have changed to use
the appropriate getter get...(..) method where relevant for relevant value
lookup.

raboczi:
> They're private scoped in JwtClaimsSet, so can't be accessed (and, hence not used in key assignment of builder)... however, I have changed to use the appropriate getter get...(..) method where relevant for relevant value lookup.
> […](#)

Yeah, I didn't notice they were private rather than part of the public interface.

fsmicdev:
Removed <keycloak.version>12.0.4</keycloak.version> property, and confirmed
built fine.

Apromore-Core-Components/Apromore-Portal/src/test/java/org/apromore/portal/security/helper/JwtHelperUnitTest.java
raboczi: Replaced testJWT with a correctly encoded version.

Apromore-Core-Components/Apromore-Portal/src/test/java/org/apromore/portal/security/helper/JwtHelperUnitTest.java
raboczi: Removed.

Apromore-Core-Components/Apromore-Portal/src/main/java/org/apromore/portal/ApromoreZKListener.java
```diff
@@ -0,0 +1,200 @@
...
+            final JWTClaimsSet jwtClaimsSet = JwtHelper.getClaimsSetFromJWT(appAuthHeader);
+
+            final String issuedAtStr = jwtClaimsSet.getStringClaim(JwtHelper.STR_JWT_KEY_ISSUED_AT);
+            LOGGER.debug("issuedAtStr {}", issuedAtStr);
+            final Date issuedAtDate = jwtClaimsSet.getIssueTime();
+            LOGGER.trace("issuedAtDate {}", issuedAtDate);
+            final Date expiryAtDate = jwtClaimsSet.getExpirationTime();
+            LOGGER.debug("expiryAtDate {}", expiryAtDate);
+
+            if (!checkDigitalSignature(appAuthHeader, signedAuthHeader)) {
```
raboczi: Checking the signature is actually a relatively expensive operation, so we'd ultimately prefer not to check it on every ZK Execution as we're currently doing here.  We should be able to optimize this in future by caching the verified JWT on the user's ZK Session and only redoing the signature check if the JWT from the App_Auth cookie is different.  For now we're proposing to just confirm that everything otherwise works, before adding such performance optimizations.

Apromore-Core-Components/Apromore-Portal/src/main/java/org/apromore/portal/dialogController/UserMenuController.java
```diff
@@ -213,17 +213,19 @@ public void onEvent(Event event) {
...
+                                                    config.getSecurityMsHttpsLogoutUrl());
+                                    LOGGER.info("\nlogoutSuccess: {}", logoutSuccess);
+                                }
                             }
 
-                            final Session session = Sessions.getCurrent();
-
                             if ((session == null || event.getData().equals(session)) || (logoutSuccess)) {
                                 Clients.evalJavaScript("window.close()");
                                 Executions.sendRedirect("/j_spring_security_logout");
```
brucenguyen1, line 231: The same comment as discussed before: this event listener is visible to all sessions, the keycloak logout doesn't check whether the received event is for this session (i.e. whether it is targeted this currentUser or not) while the redirect to Spring logout checks that with "event.getData).equals(session)".

Apromore-Core-Components/Apromore-Portal/src/main/java/org/apromore/portal/ApromoreZKListener.java
```diff
@@ -0,0 +1,200 @@
...
+        LOGGER.debug("Initialize execution {} with parent {}", exec, parent);
+
+        // If there is a parent execution, it will have already performed the required work
+        if (parent != null) {
+            return;
+        }
+
+        final HttpServletRequest httpServletRequest = (HttpServletRequest) exec.getNativeRequest();
+        final HttpServletResponse httpServletResponse = (HttpServletResponse) exec.getNativeResponse();
+
```
brucenguyen1: The main polling for JWT token (near) expiry is put under init() method of ExecutionInit, meaning it is only executed if there is an request sent from browser to the server (https://www.zkoss.org/javadoc/8.0.0/zk/org/zkoss/zk/ui/Execution.html). If the user is idle (no requests occur) and the HTTP session times out (HTTP session was automatically destroyed by web server and ZK session was destroyed by ZK), this method init() is still not executed. Now, if the user makes any action, this code here won't be executed because ZK has destroyed all desktops and UI components created before in the session (https://www.zkoss.org/wiki/ZK_Developer's_Reference/UI_Patterns/Session_Timeout_Management), the user was pointed back to the login.zul (configured in zk.xml) and asked to log in again. The state of user login on Keyloak server and any related components would be inconsistent now: they are not logged out yet because this code is not executed. Any issue (e.g. is it all the cookies related to this session are removed or still there)?

Apromore-Clients/manager-client/src/main/java/org/apromore/manager/client/ManagerServiceImpl.java
```diff
@@ -145,8 +145,11 @@ public boolean logoutUserAllSessions(final String username,
         final Properties securityMsProps = readSecurityMsProperties();
```
brucenguyen1: securityMsProps is unused.

Apromore-Core-Components/Apromore-Portal/src/main/java/org/apromore/portal/security/helper/JwtHelper.java
```diff
@@ -72,23 +85,50 @@ public static String readCookie(final HttpServletRequest httpServletRequest,
...
-        final long jwtExpiryEpoch;
-        if (expiredAtStr != null) {
-            jwtExpiryEpoch = Long.valueOf(expiredAtStr.substring(3));
-        } else { // Calculate relative
-            final Long jwtIssuedAt = Long.valueOf(issuedAtStr.substring(3));
-
-            jwtExpiryEpoch = calculateExpiryTime(jwtIssuedAt, WEBAPP_SSO_SESSION_TIMEOUT);
-        }
+                                       final Date issuedAtDate,
+                                       final Date expiryAtDate) {
```
brucenguyen1: issuedAtDate and expiryAtDate are unused.

Apromore-Core-Components/Apromore-Portal/src/main/java/org/apromore/portal/ApromoreZKListener.java
brucenguyen1: Does the use of ZK's Execution hook work in this case: if the user just accidentally or temporarily closes all the browser tabs of Apromore including the Portal for a while, then s/he opens the Portal again by typing in the URL box? This scenario happens in real life. For example, the user may open too many tabs while working (e.g. Apromore, Facebook, Youtube, news, etc.) and s/he decides to close all of them, rests for a while and then comes back to opening Apromore portal again.

Apromore-Core-Components/Apromore-Portal/src/main/java/org/apromore/portal/dialogController/UserMenuController.java
brucenguyen1: Should the processing on ZK window and redirect to Spring only be executed if "logoutSuccess = true"?

What happens if the managerService.logoutUserAllSessions fails for some reason (REST endpoint fails) and logoutSuccess = false, so the current user doesn't log out Keycloak but ZK window is closed and Spring logout is executed -> inconsistent state?

brucenguyen1: It's not recommended to put the logout logic inside UI event listener. It'd better put in a separate logic class to manage the logout processing logic, e.g. process keycloak logout, process ZK session, redirect to Spring logout, etc. It helps build an abstraction layer on top of these various misc. details; meanwhile also support unit tests.

brucenguyen1 (changes requested):
Have some comments to review.

Apromore-Core-Components/Apromore-Portal/src/main/java/org/apromore/portal/ApromoreZKListener.java
```diff
@@ -0,0 +1,200 @@
...
+    }
+
+    /**
+     * Broadcast that this session has been signed out.
+     */
+    private static void signOut(final Session session) {
+        EventQueues.lookup("signOutQueue", EventQueues.APPLICATION, true)
+                .publish(new Event("onSignout", null, session));
+    }
+}
```
brucenguyen1, line 203: Should use Session scope because the event to be published here is specific to session (it attachs session there). In addition, it's inefficient for other listeners on many other sessions to process the event not for them.

Apromore-Core-Components/Apromore-Portal/src/main/java/org/apromore/portal/dialogController/UserMenuController.java
brucenguyen1: Why doing both Clients.evalJavaScript("window.close()") and Executions.sendRedirect("/j_spring_security_logout")? The former will close the window (browser tab) and so it won't be able to show the login page in the same window.

In addition, putting redirect in this kind of event listener will lead to too many redirect errors. Imagine opening many browser tabs, each can be multiple Dashboards, multiple PDs, log animation, and so when they respond to a publish "sign out" event, there are many redirect requests.

image

Apromore-Core-Components/Apromore-Portal/src/main/java/org/apromore/portal/security/helper/JwtHelper.java
fsmicdev: Removed these two unused params.

Apromore-Clients/manager-client/src/main/java/org/apromore/manager/client/ManagerServiceImpl.java
fsmicdev: Removed unused param.

Apromore-Core-Components/Apromore-Portal/src/main/java/org/apromore/portal/ApromoreZKListener.java
nolantellis: the method name is a bit odd. Does not indicate return of boolean. , Minor change , may be rename to "isSomething"

Apromore-Core-Components/Apromore-Portal/src/main/java/org/apromore/portal/ApromoreZKListener.java
raboczi: > The state of user login on Keyloak server and any related components would be inconsistent now: they are not logged out yet because this code is not executed. Any issue (e.g. is it all the cookies related to this session are removed or still there)?

As far as session consistency goes, there are two sessions that matter:
* JSESSIONID cookie, stored by reference to an HttpSession managed by Jetty (the ZK Session wraps this)
* App_Auth/Signed_App_Auth cookies, stored by value in the JWT format (includes its own expiry time)

As long as both of these sessions are current, the user can access the system.  Reasoning by cases, it doesn't seem to matter whether the cookies are cleared or not after they expire:
*  **JSESSIONID absent:**  Jetty will generate a new JSESSIONID cookie, causing the previous ZK Session to be abandoned.  The new ZK Session doesn't have a user attribute, so they're redirected to the login page.
* **JSESSIONID present:**  The identifier it contains no longer corresponds to any HttpSession Jetty has recorded (it expired after 30 minutes of inactivity).  Jetty will generate a new HttpSession and JSESSIONID cookie, and things proceed as above since the new HttpSession doesn't have a user attribute.
* **App_Auth/Signed_App_Auth absent:**  The Execution hook fails when it doesn't find the cookie and triggers sign out.
* **App_Auth/Signed_App_Auth present:**  The JWT fails validation when its expiry time is checked, triggering sign out as before.

If the Keycloak server is managing some kind of SSO session so that you only have to enter passwords or use 2FA once, that's really its business.  My understanding is that we're not currently doing anything so fancy.  Once we've got a JWT from it, we don't care whether it remembers that our browser still exists.

Apromore-Core-Components/Apromore-Portal/src/main/java/org/apromore/portal/RefreshTokenResponse.java
```diff
@@ -0,0 +1,73 @@
...
+ * #L%
+ */
+package org.apromore.portal;
+
+public class RefreshTokenResponse {
+    private String authHeader;
+    private String signedAuthHeader;
+    private boolean wasRefreshed;
+
+    public RefreshTokenResponse() {
```
nolantellis: Can use lombok here and remove all code.

Apromore-Core-Components/Apromore-Portal/src/main/java/org/apromore/portal/dialogController/UserMenuController.java
```diff
@@ -213,17 +213,19 @@ public void onEvent(Event event) {
...
+                            final Session session = Sessions.getCurrent();
+
                             if (config.isUseKeycloakSso()) {
-                                logoutSuccess =
-                                        managerService.logoutUserAllSessions(
-                                                currentUser.getUsername(),
-                                                config.getSecurityMsHttpLogoutUrl(),
-                                                config.getSecurityMsHttpsLogoutUrl());
-                                LOGGER.info("\nlogoutSuccess: {}", logoutSuccess);
+                                if (event.getData().equals(session)) {
```
nolantellis: can we add this in a && clause of the outter if clause

Apromore-Core-Components/Apromore-Portal/src/main/java/org/apromore/portal/dialogController/TokenHandoffController.java
```diff
@@ -78,9 +77,9 @@ public void doAfterCompose(final Window window) {
                     (HttpServletRequest) Executions.getCurrent().getNativeRequest();
             LOGGER.debug("httpServletRequest: {}", httpServletRequest);
 
-            final String appAuthCookieStr = JwtHelper.readCookie(httpServletRequest, "App_Auth");
```
nolantellis, line 81: The constants defined in this calss is duplicated in jwtHelper.

Apromore-Core-Components/Apromore-Portal/src/main/java/org/apromore/portal/ApromoreZKListener.java
```diff
@@ -0,0 +1,200 @@
...
+    }
+
+    /**
+     * @param jwtStr  a plaintext JSON web token
+     * @param base64EncodedAndSignedJwtStr  an encoded and signed JSON web token
+     * @return whether  <var>base64EncodedAndSignedJwtStr</var> has content matching
+     *     <var>jwtStr</var> and is correctly encoded and signed by the key pair
+     *     identified by {@link KEY_ALIAS}
+     */
+    private boolean checkDigitalSignature(final String jwtStr, final String base64EncodedAndSignedJwtStr) {
```
nolantellis: public static boolean isSignedStrVerifiable()
y is this method in jwthelper not used.

Apromore-Core-Components/Apromore-Portal/src/main/java/org/apromore/portal/ApromoreZKListener.java
raboczi: > Does the use of ZK's Execution hook work in this case: if the user just accidentally or temporarily closes all the browser tabs of Apromore including the Portal for a while, then s/he opens the Portal again by typing in the URL box?

As long as the JSESSIONID and App_Auth/Signed_App_Auth cookies are still on their browser and they haven't delayed so long that either the HttpSession or JWT have expired, they ought to still be able to open the Portal, yes.

Apromore-Core-Components/Apromore-Portal/src/main/java/org/apromore/portal/ApromoreZKListener.java
raboczi: It looks like the onSignout event has been broken for some time; it no longer closes the user's other tabs when the user signs out.  (This feature depended on unstandardized Javascript behavior, so it might just be that browsers are more strict nowadays about letting scripts close windows.)

Apromore-Core-Components/Apromore-Portal/src/main/java/org/apromore/portal/ApromoreZKListener.java
fsmicdev: Changed to use JwtHelper method and better named method ( i.e. isJwtDigitalSignatureVerified ) in the listener class.

Apromore-Core-Components/Apromore-Portal/src/main/java/org/apromore/portal/dialogController/TokenHandoffController.java
fsmicdev: Removed unused constants from top of class.

Apromore-Core-Components/Apromore-Portal/src/main/java/org/apromore/portal/dialogController/UserMenuController.java
fsmicdev: Changed to one _if_ statement
`if ((config.isUseKeycloakSso()) && (event.getData().equals(session)))`
rather than nested _if_ statements.

Apromore-Core-Components/Apromore-Portal/src/main/java/org/apromore/portal/ApromoreZKListener.java
brucenguyen1: Your explanation above is contradictory with what you do programming. In normal case, when the user clicks "Sign Out", the event listener in UserMenuController will call to ManagerService.logoutUserAllSessions() which will call to a REST endpoint to do some logout process with Keycloak server, then calls to Spring logout process and you're trying to make these processes work consistently. In the exceptional case above, you say it's fine, the Spring logout process is done but REST endpoint logout process is not done.

Apromore-Core-Components/Apromore-Portal/src/main/java/org/apromore/portal/ApromoreZKListener.java
fsmicdev: Renamed to better named method per suggestion.

Apromore-Core-Components/Apromore-Portal/src/main/java/org/apromore/portal/RefreshTokenResponse.java
fsmicdev: Utilised Lombok, to remove boilerplate code; greatly simplified as follows:

```
@NoArgsConstructor
@AllArgsConstructor
@ToString
public class RefreshTokenResponse {

    @Getter @Setter private String authHeader;
    @Getter @Setter private String signedAuthHeader;
    @Getter @Setter private boolean wasRefreshed;
}
```

Apromore-Core-Components/Apromore-Portal/src/main/java/org/apromore/portal/dialogController/UserMenuController.java
raboczi: > Why doing both Clients.evalJavaScript("window.close()") and Executions.sendRedirect("/j_spring_security_logout")? The former will close the window (browser tab) and so it won't be able to show the login page in the same window.

It's tricky and inobvious.  The time this code was written, browsers would only allow "window.close()" to succeed for tabs that had been opened by scripted action such the BPMN editor, PD, etc.  You couldn't close a window the user had directly opened like the Portal.  By doing it this way, all the tabs except for the Portal would be closed, and the Portal would show the login page.  Otherwise, you'd end up with every tab showing a login page.

Apromore-Core-Components/Apromore-Portal/src/main/java/org/apromore/portal/ApromoreZKListener.java
raboczi: The optimization to only check signatures we haven't previously checked has been implemented.

nolantellis (approved):

brucenguyen1 (approved):

PR #1010, opened 2021-05-27
fix: SQL exception with long csv schema table when scheduling

mayankgituni:
This PR fixed the bug when the column schema info is too short for the CSV table while scheduling the pipeline.


raboczi (approved):
This builds, runs, and has been confirmed to both migrate a preexisting database or to construct a new one from an initially empty database.

PR #1018, opened 2021-06-01
AP-3892: Added import.log.maxEventCount configuration property

raboczi:
This PR allows the number of events in logs uploaded by the CSV Importer to be limited by specifying the property `import.log.maxEventCount` in site.cfg.  Logs larger than the configured limit will be truncated with the user notified.  When the property is absent, event count will be unlimited i.e. back-compatible to pre-existing site.cfg.

Note that this only applies to event log formats handled by the CSV Importer: CSV, Excel, and Parquet.  It does not limit XES or MXML.


franklma (approved):
Compile and run successfully. The log is truncated when the limitation property is set.

PR #1023, opened 2021-06-04
AP-3892: Unit tests and fixes for off-by-one upload limit errors

raboczi:
This PR adds some tests for the accuracy of the log upload limit, which previously tended to be off by one (i.e. setting the limit to 10000 would allow uploads of 10001 events).   It is a follow-up to PR https://github.com/apromore/ApromoreCore/pull/1018


franklma (approved):
Passed all tests, compile and run without issue. The off-by-one upload limit doesn't occur anymore.

PR #1024, opened 2021-06-04
AP-3925 ETL plugin: Support import of CSV/Parquet files in zip/gz

franklma:
This PR's ApromoreEE counterpart is https://github.com/apromore/ApromoreEE/pull/399


raboczi (approved):
Making MediaImpl public seems reasonable.

PR #1029, opened 2021-06-07
Remove vulnerable httpclient-4.2.1 dependency

raboczi:
This PR removes Apache HTTP client v4.2.1 source code dependencies to satisfy a  Dependabot warning about versions < 4.5.13


franklma (approved):
Compile and run without issue.

PR #1046, opened 2021-06-20
bugfix/AP-3951/DelimeterUtilMethodIncorrectResult

Janite:
This PR loosens the condition in the Delimiter class and allows the return of determined delimiter even when the columns of content is not the same as headers.

This resolves the following test failures:
- org.apromore.commons.utils.DelimiterUnitTest#testInvalidCSV
- org.apromore.service.csvimporter.services.legacy.LogImporterCSVImplUnitTest#testPrepareXesModel_test7_record_invalid
- org.apromore.service.csvimporter.services.ParquetImporterCSVImplUnitTest#testPrepareXesModel_test7_record_invalid


franklma (approved):
Work as expected.

mayankgituni (approved):

PR #1047, opened 2021-06-21
AP-4034

chiichang:
- code modified to improve performance.
- code quality can be ignored because they have been re-implemented in the upcoming version


Apromore-Extras/APMLogModule/src/main/java/org/apromore/apmlog/LaLog.java
```diff
@@ -36,10 +35,7 @@
 import org.eclipse.collections.impl.map.mutable.UnifiedMap;
 import org.eclipse.collections.impl.set.mutable.UnifiedSet;
 
-import java.util.ArrayList;
-import java.util.BitSet;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
```
raboczi: One of our Checkstyle rules is "don't use wildcarded imports".  It's not enforced by the build process...yet.  It would be better to do full import lists so there's less work to enable Checkstyle in future.

Apromore-Extras/APMLogModule/src/main/java/org/apromore/apmlog/filter/PLog.java
```diff
@@ -431,6 +433,8 @@ public void setEventSize(long eventSize) {
     public void setPTraceList(List<PTrace> pTraceList) {
         this.pTraceList.clear();
         this.pTraceList.addAll(pTraceList);
+
+
```
raboczi, line 437: Any reason to add this whitespace?

Apromore-Extras/APMLogModule/src/main/java/org/apromore/apmlog/filter/PLog.java
```diff
@@ -778,27 +787,22 @@ private PTrace getPTraceByImmutableIndex(int index) {
...
+        List<PTrace> theCusPTraceList = originalPTraceList;
 
-        for (ATrace aTrace : immutableTraces) {
-            PTrace pt = null;
-            if (!currentBS.get(aTrace.getImmutableIndex())) {
-                pt = new PTrace(mutableIndex, aTrace, this);
-                pt.setValidEventIndexBS(new BitSet(aTrace.getEventSize()));
+        for (PTrace pTrace : theCusPTraceList) {
+            if (currentBS.get(pTrace.getImmutableIndex())) {
+                pTrace.setValidEventIndexBS(idPTraces.get(pTrace.getCaseId()).get(0).getValidEventIndexBitSet());
```
raboczi, line 798: Can you guarantee that .setValidEventIndexBS is never empty, or that the IndexOutOfBounds exception will be handled when .get(0) is called?

Apromore-Extras/APMLogModule/src/main/java/org/apromore/apmlog/immutable/ImmutableLog.java
```diff
@@ -24,17 +24,14 @@
...
 import org.apromore.apmlog.stats.EventAttributeValue;
 import org.eclipse.collections.impl.bimap.mutable.HashBiMap;
 import org.eclipse.collections.impl.list.mutable.primitive.DoubleArrayList;
 import org.eclipse.collections.impl.map.mutable.UnifiedMap;
 import org.eclipse.collections.impl.set.mutable.UnifiedSet;
 
-import java.util.ArrayList;
-import java.util.List;
-import java.util.LongSummaryStatistics;
+import java.util.*;
```
raboczi, line 34: Checkstyle will complain about collapsing these imports into a wildcard.

Apromore-Extras/APMLogModule/src/main/java/org/apromore/apmlog/stats/CaseAttributeValue.java
```diff
@@ -75,7 +88,7 @@ public long getCases() {
     }
 
     public String getFrequency() {
```
raboczi, line 90: Maybe this should be renamed to "getPercentFormatted" to make it clear that it's the same value as "getPercent".

Apromore-Extras/APMLogModule/src/main/java/org/apromore/apmlog/stats/EventAttributeValue.java
```diff
@@ -82,8 +82,30 @@ public EventAttributeValue(String value, IntArrayList occurCaseIndexes, long tot
...
+                .map(AActivity::getImmutableTraceIndex)
+                .collect(Collectors.toSet());
+    }
+
+    public EventAttributeValue(String value, UnifiedSet<Integer> occurCaseIndexes, long totalCases,
+                               UnifiedSet<AActivity> occurActivities) {
+        this.value = value.intern();
+        this.occurCasesIndexSet = occurCaseIndexes;
+        this.percent = 100 * ((double) occurCaseIndexes.size() / totalCases);
+        this.frequency = String.format("%.2f", percent );
```
raboczi, line 96: I think you can remove these, since they're now calculated on demand from occurCaseIndexSet in the .getPercent and .getFrequency methods.

Apromore-Extras/APMLogModule/src/main/java/org/apromore/apmlog/stats/CaseAttributeValue.java
```diff
@@ -55,11 +57,22 @@ public CaseAttributeValue(String value, IntArrayList occurCaseIndexes, long tota
...
         this.oppCases = totalCases - occurCaseIndexes.size();
+
+        occurCaseIndexSet = Arrays.stream(occurCaseIndexes.toArray()).boxed().collect(Collectors.toCollection(UnifiedSet::new));
+    }
+
+    public CaseAttributeValue(String value, UnifiedSet<Integer> occurCaseIndexSet, long totalCases) {
+        this.value = value.intern();
+        this.occurCaseIndexSet = occurCaseIndexSet;
+        this.percent = 100 * ((double) occurCaseIndexSet.size() / totalCases);
+        this.frequency = String.format("%.2f",  percent );
```
raboczi, line 68: These two fields have been replaced by getters; I think you can remove them.

Apromore-Extras/APMLogModule/src/main/java/org/apromore/apmlog/stats/EventAttributeValue.java
```diff
@@ -132,37 +155,34 @@ public double getOppCases() {
...
-        if (occurCaseIndexes == null) {
-            int[] array = occurCasesIndexSet.stream().mapToInt(x -> x).toArray();
-            occurCaseIndexes =new IntArrayList(array);
-        }
-        return occurCaseIndexes;
+        return new IntArrayList(getOccurCasesIndexSet().stream().mapToInt(x->x).toArray());
     }
 
+    @Override
     public double getPercent() {
```
raboczi, line 162: Just like in CaseAttributeValue, I think this ought to be "getFrequency" without being a percentage, and the current "getFrequency" ought to be `formatAsPercent(getFrequency())`.

raboczi (changes requested):
I spotted some methods that I think are no longer used and should be removed.  Otherwise, my comments are only suggestions, not requirements.

One general observation is that in some places properties are stored as percentages rather than as plain fractions.  For example, CaseAttributeValue.setRatio takes a percentage instead of the actual ratio.  I feel like it would be better to treat the conversion to percentages as a presentation detail, rather than the native way data is stored.

Apromore-Extras/APMLogModule/src/main/java/org/apromore/apmlog/filter/PLog.java
raboczi: This whitespace seems spurious.

Apromore-Extras/APMLogModule/src/main/java/org/apromore/apmlog/immutable/ImmutableLog.java
raboczi: Checkstyle will complain about collapsing these imports into a wildcard.

Apromore-Extras/APMLogModule/src/main/java/org/apromore/apmlog/stats/CaseAttributeValue.java
raboczi, line 68: Fields has been replaced by getters and should be removed.

Apromore-Extras/APMLogModule/src/main/java/org/apromore/apmlog/stats/CaseAttributeValue.java
```diff
@@ -92,7 +105,7 @@ public double getOppCases() {
     }
 
     public double getPercent() {
-        return percent;
+        return 100 * ((double) getOccurCasesIndexSet().size() / totalCases);
     }
```
raboczi, line 109: The fact that .getFrequency and .getPercent are really the same value formatted different ways isn't obvious from the names.  Perhaps something like:
```
public double getFrequency() {
    return ((double) getOccurCasesIndexSet().size()) / totalCases;
}

static String formatAsPercentage(double n) {
    return String.format("%.2f", 100 * n);
}

public String getFrequencyFormattedAsPercentage() {
    return formatAsPercentage(getFrequency());
}
```

Apromore-Extras/APMLogModule/src/main/java/org/apromore/apmlog/stats/EventAttributeValue.java
raboczi, line 96: I think you can remove these, since they're now calculated on demand from occurCaseIndexSet in the .getPercent and .getFrequency methods.

Apromore-Extras/APMLogModule/src/main/java/org/apromore/apmlog/stats/EventAttributeValue.java
```diff
@@ -132,37 +155,34 @@ public double getOppCases() {
...
-        return percent;
+        return 100 * ((double) getOccurCasesIndexSet().size() / totalCases);
     }
```
raboczi, line 164: Just like in CaseAttributeValue, I think this ought to be "getFrequency" without being a percentage, and the current "getFrequency" ought to be `formatAsPercent(getFrequency())`.

raboczi (changes requested):

Apromore-Extras/APMLogModule/src/main/java/org/apromore/apmlog/stats/CaseAttributeValue.java
chiichang: they are the interfaces for ZUL.

raboczi: That's a good reason.

raboczi (approved):

PR #1048, opened 2021-06-22
Feature/current jwt map for user

fsmicdev:


Apromore-Core-Components/Apromore-Portal/src/main/java/org/apromore/portal/ApromoreZKListener.java
```diff
@@ -25,10 +25,7 @@
 import com.nimbusds.jwt.JWT;
 import com.nimbusds.jwt.JWTClaimsSet;
 import java.io.IOException;
-import java.util.Dictionary;
-import java.util.Enumeration;
-import java.util.HashMap;
-import java.util.Map;
+import java.util.*;
```
raboczi: Avoid wildcarded imports -- it's actually one of our Checkstyle rules, so theoretically we'll have to put it all back in order to enable Checkstyle on the Portal.

Apromore-Core-Components/Apromore-Portal/src/main/java/org/apromore/portal/dialogController/UserMenuController.java
```diff
@@ -28,11 +28,8 @@
 
 import java.io.UnsupportedEncodingException;
 import java.net.URLEncoder;
-import java.util.Comparator;
-import java.util.List;
-import java.util.Locale;
-import java.util.SortedMap;
-import java.util.TreeMap;
+import java.util.*;
```
raboczi, line 31: Shouldn't use import wildcards.

raboczi (approved):
Style seems fine, builds but I don't have the KC setup to functionally test it locally.

I have a comment about avoiding import wildcards, but otherwise I approve this.

Apromore-Core-Components/Apromore-Portal/src/main/java/org/apromore/portal/ApromoreZKListener.java
fsmicdev: K, will fix now.

Apromore-Core-Components/Apromore-Portal/src/main/java/org/apromore/portal/dialogController/UserMenuController.java
fsmicdev: Cool - have removed.

PR #1049, opened 2021-06-22
AP-4055 Only allow supported cloud storage URL in File Importer

franklma:
Only allow importing from supported cloud storage (Dropbox, Google Drive, Microsoft OneDrive) instead of any domains to avoid potential vulnerabilities. 

Supported Protocols: HTTPS
Supported domains: "www.dropbox.com", "drive.google.com", "onedrive.live.com"

### DIFF
--- a/Apromore-Core-Components/Apromore-Portal/src/main/java/org/apromore/portal/dialogController/ImportController.java
+++ b/Apromore-Core-Components/Apromore-Portal/src/main/java/org/apromore/portal/dialogController/ImportController.java
@@ -316,7 +316,7 @@ public class ImportController extends BaseController {
             }
             okButton_URL.setDisabled(false);
 
-        } catch (URISyntaxException | MalformedURLException e) {
+        } catch (MalformedURLException e) {
             okButton_URL.setDisabled(true);
             throw new ExceptionImport("URL link is empty or not correct.");
         } catch (IOException | NullPointerException e) {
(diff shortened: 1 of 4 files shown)


Apromore-Core-Components/Apromore-Portal/src/main/java/org/apromore/portal/dialogController/ImportController.java
```diff
@@ -308,7 +316,7 @@ private void uploadFileFromURL(String fileUrl) throws ExceptionImport, Exception
             }
             okButton_URL.setDisabled(false);
 
-        } catch (MalformedURLException e) {
+        } catch (URISyntaxException | MalformedURLException e) {
```
nolantellis: catch this inside isValidCloudStorageURL log it and return false and through the

Apromore-Core-Components/Apromore-Portal/src/main/java/org/apromore/portal/util/StringUtil.java
```diff
@@ -44,10 +45,13 @@
...
             '=', ','};
     private static final char SANITIZED_CHAR = '_';
+    private static final int FILE_NAME_MAX_LENGTH = 255;
 
-    private static final String dropboxPrefix = "https://www.dropbox.com";
-    private static final String googleDrivePrefix = "https://drive.google.com";
-    private static final String oneDrivePrefix = "https://onedrive.live.com";
+    private static final String VALID_PROTOCOL = "https://";
+
+    private static final String DROPBOX_DOMAIN = "www.dropbox.com";
```
nolantellis, line 52: add this to a list and check with contains

nolantellis (changes requested):

Apromore-Core-Components/Apromore-Portal/src/main/java/org/apromore/portal/util/StringUtil.java
franklma: The reason to have separate domains here is they are used in other places to identify URL of different domains and convert from sharable URL to downloadable URL accordingly.

nolantellis (approved):

PR #1052, opened 2021-06-23
AP-3474: Changes to logging

raboczi:
This PR progresses towards making the system logs useful for system administrators.  The intention is that log messages filtered at the INFO level will show only high-level user activities and which artifacts the users are interacting with.  Other details of interest solely for development or debugging should be at DEBUG level.
* Many log messages have been reduced from INFO level to DEBUG.
* User login and logout (whether explicit or via session timeout) should now be logged.
* System.out and e.printstackTrace() etc have been replaced with logging methods.
* Many log methods have being adjusted to use a template-and-subsitution-list style rather than string concatenation.


Apromore-Core-Components/Apromore-Manager/src/main/java/org/apromore/service/impl/ProcessServiceImpl.java
```diff
@@ -354,9 +354,9 @@ public ExportFormatResultType exportProcess(final String name, final Integer pro
             org.springframework.security.core.Authentication auth =
                 org.springframework.security.core.context.SecurityContextHolder.getContext().getAuthentication();
             if (auth != null) {
-                LOGGER.info("Authentication principal=" + auth.getPrincipal() + " details=" + auth.getDetails() + " thread=" + Thread.currentThread());
+                LOGGER.debug("Authentication principal={} details={} thread={}", auth.getPrincipal(), auth.getDetails(), Thread.currentThread());
```
brucenguyen1: This must be LOGGER.info, not debug level.

Apromore-Core-Components/Apromore-Manager/src/main/java/org/apromore/service/impl/ProcessServiceImpl.java
```diff
@@ -354,9 +354,9 @@ public ExportFormatResultType exportProcess(final String name, final Integer pro
...
             } else {
-                LOGGER.info("Authentication is null");
+                LOGGER.debug("Authentication is null");
```
brucenguyen1: This must be LOGGER.error, not debug because null authentication is an exceptional condition (a type of error) and it must be highlighted in the log.

Apromore-Core-Components/Apromore-Manager/src/main/java/org/apromore/service/impl/ProcessServiceImpl.java
```diff
@@ -504,25 +504,25 @@ public String getBPMNRepresentation(final String name, final Integer processId,
             org.springframework.security.core.Authentication auth =
                     org.springframework.security.core.context.SecurityContextHolder.getContext().getAuthentication();
             if (auth != null) {
-                LOGGER.info("Authentication principal=" + auth.getPrincipal() + " details=" + auth.getDetails() + " thread=" + Thread.currentThread());
+                LOGGER.debug("Authentication principal={} details={} thread={}", auth.getPrincipal(), auth.getDetails(), Thread.currentThread());
             } else {
-                LOGGER.info("Authentication is null");
+                LOGGER.debug("Authentication is null");
```
brucenguyen1: Same comment as above.

franklma (approved):
Compile and run without any issue. INFO and DEBUG loggers work as expected.

Apromore-Custom-Plugins/Log-Animation-Portal-Plugin/src/main/java/org/apromore/plugin/portal/loganimation/LogAnimationCleaner.java
```diff
@@ -40,7 +40,7 @@
 
     @Override
     public void cleanup(Desktop desktop) throws Exception {
-        LOGGER.info("LogAnimation cleanup starts for desktopID = " + desktop.getId());
+        LOGGER.debug("LogAnimation cleanup starts for desktopID = {}", desktop.getId());
         
         // Clean up this plugin
         // Log animation logic has no tricky resources to be cleaned up, auto cleaned by the GC
```
brucenguyen1: This should be INFO since it marks the end of the plugin execution. Pls do not decide on what is done here (memory cleanup or whatever), this marks the end of the plugin and it needs some logging. Memory or something cleanup to do  only uses this exit point to do the job.

Apromore-Custom-Plugins/Process-Discoverer-Portal-Plugin/src/main/java/org/apromore/plugin/portal/processdiscoverer/PDDesktopCleaner.java
```diff
@@ -39,13 +39,13 @@
 
     @Override
     public void cleanup(Desktop desktop) throws Exception {
-        LOGGER.info("PD cleanup starts for desktopID = " + desktop.getId());
+        LOGGER.debug("PD cleanup starts for desktopID = " + desktop.getId());
```
brucenguyen1: This should be INFO level as it marks the end of the plugin execution.

brucenguyen1 (changes requested):

Apromore-Custom-Plugins/Log-Animation-Portal-Plugin/src/main/java/org/apromore/plugin/portal/loganimation/LogAnimationCleaner.java
raboczi: This is precisely the spammy, non-user-initiated internal detail that we're trying to remove from INFO level.  It needs to be logged, and it gets logged at DEBUG level.  The system administrator does not want to see session management details like this.  They just care that the user logged out, and that gets logged at INFO level.

Apromore-Core-Components/Apromore-Manager/src/main/java/org/apromore/service/impl/ProcessServiceImpl.java
raboczi: If you look at the code block, it's explicitly commented as "Debug tracing of the authenticated principal".  The value `auth` is not used anywhere in the code, other than to be logged.  It's probably just a leftover from someone (likely Past Simon) trying to figure out whether the authenticated principle was accessible from that part of the code, before we had UserSessionManager to do that.

Would it be satisfactory to remove this logging message entirely?

Apromore-Core-Components/Apromore-Manager/src/main/java/org/apromore/service/impl/ProcessServiceImpl.java
raboczi: This is explicitly marked with a comment as "Debug tracing of the authenticated principal" which is pretty solid evidence it was always intended to be debugging.  I propose to remove it rather than make it INFO level.  Agreeable?

Apromore-Core-Components/Apromore-Manager/src/main/java/org/apromore/service/impl/ProcessServiceImpl.java
raboczi: Same proposal to remove as above.

Apromore-Custom-Plugins/Process-Discoverer-Portal-Plugin/src/main/java/org/apromore/plugin/portal/processdiscoverer/PDDesktopCleaner.java
raboczi: As above, this is not helpful for the system administrator to see.

Apromore-Core-Components/Apromore-Manager/src/main/java/org/apromore/service/impl/ProcessServiceImpl.java
raboczi: Removed.

Apromore-Core-Components/Apromore-Manager/src/main/java/org/apromore/service/impl/ProcessServiceImpl.java
raboczi: Removed.

Apromore-Custom-Plugins/Process-Discoverer-Portal-Plugin/src/main/java/org/apromore/plugin/portal/processdiscoverer/PDDesktopCleaner.java
raboczi: Reinstated to INFO level.

Apromore-Custom-Plugins/Log-Animation-Portal-Plugin/src/main/java/org/apromore/plugin/portal/loganimation/LogAnimationCleaner.java
raboczi: Reinstated at INFO level.

brucenguyen1 (approved):

PR #1053, opened 2021-06-24
Add helper method for CustomCalendar in EventLogService

franklma:


Apromore-Core-Components/Apromore-Manager/src/main/java/org/apromore/service/impl/EventLogServiceImpl.java
```diff
@@ -490,4 +490,9 @@ public Long getCalendarIdFromLog(Integer logId) {
 	return calendar == null ? 0 : calendar.getId();
     }
 
+    @Override
+    public CustomCalendar getCalendarFromLog(Integer logId) {
+        return logRepo.findUniqueByID(logId).getCalendar();
```
nolantellis: Need minor change here.  check getCalendar ==null , else get the ID and call calendarService.getCalendar(id), This will return CalendarModel, which Bruce and other will use

nolantellis (changes requested):

nolantellis (approved):

PR #1064, opened 2021-06-30
AP-4097 Spring cloud config in apromore

justinapromore:


nolantellis (commented):
This is good, but we are not ready on this front. We will use spring cloud config in CI CD. 
We can close both the review and later consider this in springboot.
For now, The intentionis to use spring cloud config in CI CD for any app, not just spring boot

PR #1070, opened 2021-07-07
AP-4155/add etl enable flag in site properties

Janite:
This PR has a pair in ApromoreEE: https://github.com/apromore/ApromoreEE/pull/433

Added an etl.enable flag in site.properties and used it to show/hide the etl + job scheduler buttons in portal.
ETL is disabled by default.


raboczi (approved):
This is straightforward and follows the pattern of previous code, so seems entirely safe.

PR #1076, opened 2021-07-09
AP-4125: Close streams

raboczi:
This PR addresses issues raised by SonarCloud regarding streams which weren't reliably being closed, particularly in case of exceptions interrupting the normal flow of execution.

### DIFF
--- a/Apromore-Custom-Plugins/CSVImporter-Logic/src/main/java/org/apromore/service/csvimporter/io/ParquetLocalFileReader.java
+++ b/Apromore-Custom-Plugins/CSVImporter-Logic/src/main/java/org/apromore/service/csvimporter/io/ParquetLocalFileReader.java
@@ -51,9 +51,9 @@ public class ParquetLocalFileReader {
             try {
                 thread.setContextClassLoader(Path.class.getClassLoader());
                 InputFile inputFile = HadoopInputFile.fromPath(new Path(file.toURI()), conf);
-                ParquetFileReader parquetFileReader = ParquetFileReader.open(inputFile);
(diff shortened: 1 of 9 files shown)


Apromore-Custom-Plugins/CSVImporter-Logic/src/main/java/org/apromore/service/csvimporter/io/ParquetLocalFileReader.java
```diff
@@ -51,9 +51,9 @@ public ParquetLocalFileReader(Configuration conf, File file) throws IOException
             try {
                 thread.setContextClassLoader(Path.class.getClassLoader());
                 InputFile inputFile = HadoopInputFile.fromPath(new Path(file.toURI()), conf);
-                ParquetFileReader parquetFileReader = ParquetFileReader.open(inputFile);
-                schema = parquetFileReader.open(inputFile).getFooter().getFileMetaData().getSchema();
-                parquetFileReader.close();
+                try (ParquetFileReader parquetFileReader = ParquetFileReader.open(inputFile)) {
+                    schema = parquetFileReader.getFooter().getFileMetaData().getSchema();
```
raboczi, line 55: Everything else in this PR is fairly pure refactoring and doesn't change the original behavior, except for this line. I'm pretty sure it wasn't intentional to call `ParquetFileReader.open` twice.

franklma:
Compile and run without issue. File importer works as expected. The try-with-resources statements should be preferred over try/catch/finally.

franklma (approved):

PR #1091, opened 2021-07-12
Update README.md

shefali316:


raboczi (changes requested):
Will need to update that the default is now MySQL rather than H2.

raboczi:
There is an out-of-date reference to port 9000 which should be changed to 8181.

raboczi (approved):
This fine to merge.  I have some comments about additional updates.

PR #1095, opened 2021-07-13
AP-4169: XML External Entity processing (XXE) vulnerabilities

raboczi:
This PR configures assorted and scattered XML processors to disable features that might allow a maliciously-crafted XML document to use entity expansion or inclusion to cause trouble.  See the relevant [OWASP cheat sheet](https://cheatsheetseries.owasp.org/cheatsheets/XML_External_Entity_Prevention_Cheat_Sheet.html#java) for additional detail.  AFAICT I haven't broken any existing functionality, but XML processing is fundamental enough to everything else that it's certainly a risk.  This ought to remove 10 issues at the Blocker / Vulnerability level from the SonarCloud report.

### DIFF
--- a/Apromore-Custom-Plugins/Apromore-Editor/bpmn20xmlbasic/src/main/java/de/unihannover/se/infocup2008/bpmn/dao/ERDFDiagramDao.java
+++ b/Apromore-Custom-Plugins/Apromore-Editor/bpmn20xmlbasic/src/main/java/de/unihannover/se/infocup2008/bpmn/dao/ERDFDiagramDao.java
@@ -66,6 +66,7 @@ import org.w3c.dom.traversal.TreeWalker;
 import org.xml.sax.InputSource;
 import org.xml.sax.SAXException;
 
+import javax.xml.XMLConstants;
 import javax.xml.parsers.DocumentBuilder;
 import javax.xml.parsers.DocumentBuilderFactory;
 import javax.xml.parsers.ParserConfigurationException;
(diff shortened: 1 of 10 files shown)


nolantellis (approved):

Apromore-Custom-Plugins/Apromore-Editor/bpmn20xmlbasic/src/main/java/de/unihannover/se/infocup2008/bpmn/dao/ERDFDiagramDao.java
```diff
@@ -339,10 +337,8 @@ public void saveToWriter(Writer writer) {
...
 
         // Write the DOM document to the file
-        Transformer xformer;
         try {
-            xformer = TransformerFactory.newInstance().newTransformer();
-            xformer.transform(source, result);
+            transform(source, result);
         } catch (TransformerConfigurationException e) {
             // TODO Auto-generated catch block
             e.printStackTrace();
```
karthik-balasubramanian, line 344: Can we please log this exception instead of printing the stack trace.

raboczi: That's a good catch, and I would've fixed it if the comment had come before it was merged into development.  Fortunately this is dead code, since we don't use ERDF from this library, only BPMN.

PR #1096, opened 2021-07-13
Hotfix/cache apmlog

franklma:


Apromore-Core-Components/Apromore-Manager/src/main/java/org/apromore/service/impl/TemporaryCacheService.java
```diff
@@ -136,7 +136,7 @@ public Storage storeProcessLog(final Integer folderId, final String logName, XLo
 		if (shouldCache(log)) {
 		    // Store corresponding object into cache
 		    cacheRepo.put(name, log);
-		    cacheRepo.put(logNameId + APMLOG_CACHE_KEY_SUFFIX, apmLogService.findAPMLogForXLog(log));
+		    cacheRepo.put(name + APMLOG_CACHE_KEY_SUFFIX, apmLogService.findAPMLogForXLog(log));
```
brucenguyen1: To be safe, the code must handle the possible null returning result from apmLogService.findAPMLogForXLog(log) which may be caused by empty logs after validation or runtime error. Otherwise, null is stored in cache leading to possible errors in later stage of cache reading. The same comment to other places calling apmLogService.findAPMLogForXLog(log).

brucenguyen1: It would help consistent and manageable code to have a common method called getCacheKey() to create cache key from folder id (maybe), log name and current date (this rule of key creation is a bit complex worthing to be managed properly). This method is to be called by the three methods storeProcessLog, getProcessLog and deleteProcessLog rather than duplicate code in those methods. It would help if any changes to do the cache key in the future as well, avoid errors caused by forgotten change in some places.

brucenguyen1:
This class should be refactored in another PR because of the duplication issues between different methods: storeProcessLog, getProcessLog, getAggregatedLog.

franklma:
Sure. will create another PR for the rest of refactoring.

brucenguyen1 (approved):

PR #1115, opened 2021-07-23
Configuration for SonarCloud (AP-4308)

raboczi:
This PR configures the build process to include SonarCloud static analysis.


raboczi:
This needs to be done differently due to the migrations from Maven to Gradle.

PR #1125, opened 2021-07-28
Removed unused Karaf and applet-related sources.

raboczi:
This PR removes the core-assemblies directory which was used to build the open source Karaf server, which doesn't work with the new build system.  It also removes a lingering applet configuration files, years after browsers stopped supporting applets or the codebase included any.


nolantellis (approved):
Thank you Simon

PR #1128, opened 2021-07-28
Redoing AP-4308 for Gradle instead of Maven.  (AP-4308)

raboczi:
This PR configures GitHub to build new PRs and scan them with SonarCloud.


raboczi:
ApromoreCore doesn't currently build under gradle, so deferring this until I have a base commit to work from.

PR #1133, opened 2021-07-28
Added dependency in core

nolantellis:


raboczi (approved):
Confirmed that ApromoreCore can be built.  Process Discoverer does not work, but I'm approving this anyway because building it is enough to unblock other work.

PR #1137, opened 2021-07-28
bugfix/AP-4348/Show log name in filter window title when accessing filter via PD

Janite:
Previously, when creating a LogFilterRequest for the filter via PD, the PD window title (blank string) was being passed in the log name field instead of the log name. Made changes to pass the log name instead.


nolantellis (approved):
Thanks Jane for addressing this

PR #1152, opened 2021-08-02
AP-2062: simplify PortalPluginResourceServlet

raboczi:
This PR simplifies PortalPluginResourceServlet, rather than removing it as originally intended by the Jira issue.  Spring Boot doesn't have any obvious mechanism for components to add static content, so doing it via a servlet is still the lesser evil.  Since there is now a single classloader, the resource path no longer needs to explicitly communicate which component it's intended for.  This means a URL like `/portalPluginResource/upload.svg` is now sufficient, rather than `/portalPluginResource/File/Upload/upload.svg`.  Other than shorter URLs, behavior should be unchanged.

### DIFF
--- a/Apromore-Core-Components/Apromore-Portal/src/main/java/org/apromore/portal/dialogController/MenuController.java
+++ b/Apromore-Core-Components/Apromore-Portal/src/main/java/org/apromore/portal/dialogController/MenuController.java
@@ -26,8 +26,6 @@
 
 package org.apromore.portal.dialogController;
 
-import java.io.UnsupportedEncodingException;
-import java.net.URLEncoder;
 import java.util.Comparator;
 import java.util.List;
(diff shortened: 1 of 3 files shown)


dreamind (approved):
Just wondering if we need to extra strict by using URLEncoder.encode() for the iconPath.

raboczi:
> Just wondering if we need to extra strict by using URLEncoder.encode() for the iconPath.

That's a reasonable idea.  I'll do it.

raboczi:
> > Just wondering if we need to extra strict by using URLEncoder.encode() for the iconPath.
> 
> That's a reasonable idea. I'll do it.

On second thoughts, no: it can't be encoded, because we'd like to be able to include path separators in the locations of icons (e.g. FilterEE/icon.svg).  I'll commit it as-is.

PR #1165, opened 2021-08-03
Ap 4233 sonar critical issue fixes in Apromore Portal

justinapromore:


brucenguyen1:
Agree our UIController classes extend from ZK's Window class which eventually extends ZK's AbstractComponent which implements Serializable interface. Note that ZK is a _generic_ UI framework, it has its AbstractComponent implements Serializable interface which allows other UI elements such as button, list, checkbox, etc. to be serialized, it is to serve various use cases in various software applications that use ZK, i.e. ZK has its own business. For our business, we need to ask a question: do we need ZK to serialize our Window, e.g. this is for the purpose of persisting Window UI and it will show up the same UI state when the user opens it the next time. 

Do we use that facility (if any) provided by ZK to serialize the whole Window? What exactly Window in Apromore we need that UI state persistence capability (definitely not all the Controller classes)? And if you envision that it's a yes, what Jira card that requests this capability from a business requirement perspective (developer's wish alone is not valid enough)?

I would implement this "transient" marker if we have the a business need that needs serialization for our UI, and it entails lots of more work to do rather than only marking transient or non-transient.

Apromore-Core-Components/Apromore-Portal/src/main/java/org/apromore/portal/dialogController/BaseController.java
```diff
@@ -143,6 +143,7 @@ protected String getURL(final String nativeType) {
       case EPML_2_0:
         url = "http://b3mn.org/stencilset/epc#";
         break;
```
nolantellis, line 145: if you have made changes here. It could be better to use url variable and return it at the end.
Rather thsan default everywhere

Apromore-Core-Components/Apromore-Portal/src/main/java/org/apromore/portal/dialogController/BaseController.java
```diff
@@ -166,6 +167,7 @@ public String getImportPath(final String nativeType) {
       case EPML_2_0:
         importPath = "/" + getConfig().getSiteEditor() + "/editor/epmlimport";
         break;
+      default:
     }
```
nolantellis, line 171: Same as above

nolantellis (approved):

PR #1171, opened 2021-08-05
Delete unused files

nolantellis:


franklma (approved):
compile and run without issue

PR #1177, opened 2021-08-06
AP-4236 Refactor menu

dreamind:
This branch is intended for menu refactoring. Must be build with the corresponding EE branch. The changes include provide an ID to each plugin, instead of relying on label (which changes based on language). Plugin is accessible via a portal plugin map and the corresponding ID. Menu structure is configurable via menu.json, instead of implicitly via group and application.properties menu order.


Apromore-Boot/src/main/resources/menu.json
```diff
@@ -0,0 +1,93 @@
+{
+  "groups": [
+    {
+      "id": "FILE",
+      "items": [
+        {
+          "id": "UPLOAD"
+        },
```
brucenguyen1: Menu is part of the Portal (it is the Portal menu). Putting it in another module like Apromore-Boot instead of the Portal module adds to the complexity of the system, a phenomenon called "Unknown Unknowns": i.e. you don't know where the things (you don't know) are, how it works, how to modify when needed, etc. 
Suggest put this menu.json file in the Portal under the same package of all menu related code.

Apromore-Core-Components/Apromore-Portal/src/main/java/org/apromore/portal/common/Constants.java
```diff
@@ -94,9 +94,37 @@
...
-    public static final String USER_ADMIN_PLUGIN = "USER_ADMIN_PLUGIN";
-    public static final String ACCESS_CONTROL_PLUGIN = "ACCESS_CONTROL_PLUGIN";
-    public static final String ETL_PLUGIN = "ETL_PLUGIN";
-    public static final String JOB_SCHEDULER_PLUGIN = "JOB_SCHEDULER_PLUGIN";
-
+    public static final String PLUGIN_USER_ADMIN = "USER_ADMIN";
+    public static final String PLUGIN_ACCESS_CONTROL = "ACCESS_CONTROL";
+    public static final String PLUGIN_ETL = "ETL";
+    public static final String PLUGIN_JOB_SCHEDULER = "JOB_SCHEDULER";
+    public static final String PLUGIN_CALENDAR = "CALENDAR";
```
brucenguyen1: This Constants.java is becoming too big with different types of constants all live next to each other here. It is becoming hard to read, differentiate or modify safely different types of constants while they look quite similar. Suggest creating a new class called MenuLabelConstants to put many menu label constants. This reduces the dependency from other classes to this class only for menu related, everything will depend on Constants. java which should be avoided.

Apromore-Boot/src/main/resources/menu.json
dreamind: Thanks, Bruce. Initially, I thought that the configuration should be configurable in the runtime and hence I put it in the ApromoreBoot. But I agree with your point that we probably should have default in the portal as well.

Apromore-Core-Components/Apromore-Portal/src/main/java/org/apromore/portal/config/PortalMenuConfig.java
```diff
@@ -0,0 +1,89 @@
...
+				for (int j = 0; j < items.length(); j++) {
+					JSONObject item = items.getJSONObject(j);
+					String itemId = item.getString("id");
+					menuItems.add(itemId);
+				}
+				menuiItemMap.put(groupId, menuItems);
+			}
+		} catch (Exception e) {
+			LOGGER.info("Failed to load menu definition");
+		}
```
brucenguyen1: The exception handling here is not complete. A question here is what is the behavior of the software when different exceptions occur in loading menu config: missing config file, IO problem, missing menu items, wrong value type, wrong format. Some are more serious than the others. 
For example: 
- Missing the menu config file is serious because it can't go forward (nothing can be shown). In this case, the program should inform the user about the error to fix and then stop.
- Missing a menu item, wrong value, etc. is less serious: the program should continue showing other valid menu items, but showing a notification for the user to know.
- Other individual issues: it should be able to continue.

For each type of exceptions above, throw a type of exception for the calling program to know and handle (here is the Portal MenuController and/or MainController). Preferred unchecked exceptions rather than checked exception to keep the API clean.

Apromore-Core-Components/Apromore-Portal/src/main/java/org/apromore/portal/dialogController/MenuController.java
```diff
@@ -80,159 +84,120 @@ public void onEvent(Event event) throws Exception {
...
-          @Override
-          public void onEvent(Event event) throws Exception {
-            PortalContext portalContext =
+  private void addPluginToMenu(Menupopup popup, String itemId) {
+    PortalPlugin plugin = portalPluginMap.get(itemId);
+    if (plugin == null) {
+      return;
+    }
+    PortalPlugin.Availability availability = plugin.getAvailability();
+    if (availability == PortalPlugin.Availability.UNAVAILABLE || availability == PortalPlugin.Availability.HIDDEN) {
```
brucenguyen1: Tab space should set to 4 (coding standards as we agreed).

Apromore-Core-Components/Apromore-Portal/src/main/java/org/apromore/portal/dialogController/MenuController.java
```diff
@@ -80,159 +84,120 @@ public void onEvent(Event event) throws Exception {
...
+            if (!(popup.getLastChild() instanceof Menuseparator)) {
+              Menuseparator separator = new Menuseparator();
+              popup.appendChild(separator);
             }
-          } catch (Exception e) {
-            LOGGER.error("Ignored exception during main menu construction", e);
+          } else {
+            addPluginToMenu(popup, itemId);
           }
         }
```
brucenguyen1: The Cut, Copy, Paste can be managed consistently under the same PortalPluginResolver mechinism (via PortalPlugin interface and beans). They are different from other PortalPlugin such as PD or Calendar in the sense that they are _internal_ PortalPlugin, i.e. their action could be handled internally by classes within the Portal module itself. 

Suggest to create a package called plugins inside the Portal, then create one plugin class for each editing action, e.g. CutPortalPlugin, CopyPortalPlugin, PastePortalPlugin which implements PortalPlugin interface, and implement the current interface such as execute. They can be loaded via PortalPluginResolver, no need to create conditionals/exceptional cases for these plugins which add another layer of complexity to manage: it requires to read these conditionals and modify this big chunk of code whenever something needs to change

Apromore-Core-Components/Apromore-Portal/src/main/java/org/apromore/portal/config/PortalMenuConfig.java
```diff
@@ -0,0 +1,89 @@
...
+import java.util.Map;
+
+import lombok.Getter;
+
+import org.apromore.plugin.portal.PortalLoggerFactory;
+
+@Component("menuConfig")
+public class PortalMenuConfig {
+
+	private static final Logger LOGGER = PortalLoggerFactory.getLogger(PortalMenuConfig.class);
```
brucenguyen1: This PortalMenuConfig needs a unit test with different types of menu available, test the exception handling as well.

Apromore-Boot/src/main/resources/menu.json
```diff
@@ -0,0 +1,93 @@
...
+        {
+          "id": "DISCOVER_MODEL"
+        },
+        {
+          "id": "EDIT_MODEL"
+        },
+        {
+          "id": "FILTER_LOG"
+        }
```
brucenguyen1: This menu.json is a good refactoring step, it should be utilized further to remove the current complex dependencies while it's being used here:
- Add the group label and menu item label to menu.json.
For example: 
{
  "id": "DISCOVER",
  "label": "Discover" (or label code to support internationalization),
   "items": [
       {
            "id": "CREATE_MODEL",
            "label": "Create Model" (or label code to support internationalization)
       }
   ]  
}
- Remove getGroup and getLabel in PortalPlugin
- Removal getGroup, getLabel, getGroupLabel... in each concrete PortalPlugin (e.g. CalendarPlugin, PD, etc.). They are decided based on the item id and menu.json
- Further, the item id such as "id": "CREATE_MODEL" can be made more generic by using the plugin class canonical name for id, e.g. "id" : "org.apromore.plugin.portal.file.CreateProcessModelPlugin". This will help remove the getID() method implementation in each concrete plugin like CreateProcessModelPlugin, just use one default implementation in PortalPlugin for getID(), e.g. this.getClass().getCanonicalName(). So, when the MenuController processes menu.json, it will match the "id" in each item of menu.json with the getID() in each bean in PortalPluginResolver. If matched, it knows which group label and menu item label to be used for that menu item. 
Another benefit is it is explicit to specify which concrete plugin class is being called for this menu item, removing the obscurity of keeping matched the ID is set in menu.json and the getID() {return Constants.CREATE_MODEL} in CreateProcessModelPlugin (you have to search hundreds of classes for which one has this value set in its getID), an "unknown unknowns" phenomenon.

Apromore-Core-Components/Apromore-Portal/src/main/java/org/apromore/portal/dialogController/MenuController.java
dreamind: I will implement the internal plugins for cut, copy, paste in a separate ticket

nolantellis:
I am a bit in a puzzle if this should go in.
I agree with the comments. 
I would prefer to convert cout/paste /copy to plugins first so that the json will remain consistent. 
Onces thts done. I can advise on the menu changes. Where we can keep the labels and names in jspon and only availability in the plugin.

dreamind:
> I am a bit in a puzzle if this should go in.
> I agree with the comments.
> I would prefer to convert cout/paste /copy to plugins first so that the json will remain consistent.
> Onces thts done. I can advise on the menu changes. Where we can keep the labels and names in jspon and only availability in the plugin.

I agree, the cut/paste/copy handling in the MenuController is rather awkward and need to be abstracted. But the issue is that having these as plugins in the way of the current file-portal-plugin seems to be a long detour since the logic still resides in the Apromore-Portal. In that context, the plugins are simply empty shells just for to add the portal internal functionalities. It looks like additional baggage of couplings. I'm wondering if we can differentiate them (file-portal plugin & copy/cut/paste) as "internal-plugins" within the portal so both the logic and the menu interfacing reside in the same module (cc @brucenguyen1 and @raboczi for comments)

nolantellis:
Spoke to bruce. After some discussion, lets add this to development and we can change if required later.

nolantellis:
pls check the merge conflicts and build and run. And then merge.

nolantellis (approved):

PR #1184, opened 2021-08-09
Misc changes for calendar,junit and pd

nolantellis:


franklma (approved):
Compile and run without issue

PR #1188, opened 2021-08-09
AP-4232 Remove exceptions from the popups during log import

franklma:


chiichang (approved):
build&run without problem

PR #1208, opened 2021-08-13
AP-4482 Support SSO group handling in account management

dreamind:


Apromore-Core-Components/Apromore-Portal/src/main/java/org/apromore/portal/common/zk/ComponentUtils.java
```diff
@@ -22,28 +22,21 @@
...
-            sclass = sclass + " " + sclassOn;
-        } else {
-            sclass = sclass.replace(sclassOn, "");
-            sclass = sclass + " " + sclassOff;
-        }
+        String oldClass = (newState) ? sclassOff : sclassOn;
+        String newClass = (newState) ? sclassOn : sclassOff;
+        sclass = sclass.replaceAll(oldClass, "");
+        sclass = sclass + " " + newClass;
         comp.setSclass(sclass);
```
raboczi: I'm a bit nervous about the string handling on sclass -- what if oldClass is a substring of something else in the sclass list, etc?  Suggest something similar to the following:
```
import java.util.Arrays;
import java.util.stream.Collectors;
import java.util.stream.Stream;

sclass = Stream.concat(Arrays.stream(sclass.split("\\s+")).filter(s -> !s.equals(oldClass)),
                       Stream.of(newClass)))
               .distinct()
               .collect(Collectors.joining(" "));
```

raboczi (changes requested):

raboczi (approved):

PR #1211, opened 2021-08-14
AP-4482 Support SSO group handling in account management

dreamind:


dreamind:
@raboczi Apparently sonar doesn't like the presence of anonymous predicate and suggested conversion to lambda :)

raboczi (approved):

PR #1213, opened 2021-08-16
bugfix/AP-4282/download-bpmn-pdf-with-name-diagram

Janite:
Updated the logic for download pdf in BPMN editor.

1. Modified AlternativesRenderer to return the PDF straight way (through the http response) instead of writing to temp folder and returning a url to the file in temp.
2. Converted the prototype js Ajax request to a XMLHttpRequest so that `responseType = 'arraybuffer'` could be added to the request. Otherwise the http response is not returned as the correct type.

### DIFF
--- a/Apromore-Core-Components/Apromore-BPMNEditor/src/main/scripts/plugins/pdf.js
+++ b/Apromore-Core-Components/Apromore-BPMNEditor/src/main/scripts/plugins/pdf.js
@@ -113,17 +113,13 @@ Apromore.Plugins.File = Clazz.extend({
 
         xhr.open("POST", Apromore.CONFIG.PDF_EXPORT_URL);
         xhr.setRequestHeader('Content-type', 'application/x-www-form-urlencoded');
-        xhr.responseType = 'arraybuffer';
+        xhr.responseType = 'blob';
         xhr.onreadystatechange = function() {
             if (xhr.readyState == 4 && xhr.status == 200) {
                 // Download pdf from blob
                 myMask.hide();
-                var resultByte = xhr.response;
-                var bytes = new Uint8Array(resultByte); // convert to byte array
-                var blob = new Blob([bytes], {type: "application/pdf"});// convert to a blob
-
                 var hiddenElement = document.createElement('a');
-                hiddenElement.href = window.URL.createObjectURL(blob);
+                hiddenElement.href = window.URL.createObjectURL(xhr.response);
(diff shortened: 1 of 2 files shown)


Apromore-Core-Components/Apromore-BPMNEditor/src/main/scripts/plugins/pdf.js
```diff
@@ -108,26 +108,33 @@ Apromore.Plugins.File = Clazz.extend({
...
-                // location.href = request.responseText
-                window.open(request.responseText);
-            }).bind(this),
-            onFailure: (function(){
+        var xhr = new XMLHttpRequest();
+        var params = "resource=" + resource + "&data=" + svgClone + "&format=pdf";
+
+        xhr.open("POST", Apromore.CONFIG.PDF_EXPORT_URL);
+        xhr.setRequestHeader('Content-type', 'application/x-www-form-urlencoded');
+        xhr.responseType = 'arraybuffer';
```
dreamind: Probably try "blob" directly as responseType, since it's tiny-bit more supported https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/responseType

Apromore-Core-Components/Apromore-BPMNEditor/src/main/scripts/plugins/pdf.js
```diff
@@ -108,26 +108,33 @@ Apromore.Plugins.File = Clazz.extend({
...
-            }).bind(this)
-        });
+                var resultByte = xhr.response;
+                var bytes = new Uint8Array(resultByte); // convert to byte array
+                var blob = new Blob([bytes], {type: "application/pdf"});// convert to a blob
+
+                var hiddenElement = document.createElement('a');
+                hiddenElement.href = window.URL.createObjectURL(blob);
+                hiddenElement.target = '_blank';
+                hiddenElement.download = 'diagram.pdf';
```
dreamind, line 124: Could we get model name?

dreamind (approved):

Apromore-Core-Components/Apromore-BPMNEditor/src/main/scripts/plugins/pdf.js
Janite: I'm having trouble figuring out how to get the model name in pdf.js. We can leave it as diagram.pdf for now and update with the model name at a later stage.

PR #1227, opened 2021-08-20
Fix unit test: fix PD Portal Plugin test to be timezone neutral

brucenguyen1:


mayankgituni (approved):
Works on my machine now!

PR #1228, opened 2021-08-22
chore: remove gradle temp files

dreamind:


nolantellis (approved):
Thank you for doing this

PR #1231, opened 2021-08-23
feature/AP-4335-add-case-variant-inspector

Janite:
Added a case variant inspector, which functions similar to the case inspector but with the following differences:

- The case variant inspector window shows the following: Case variant id, activity instances, average duration, number of cases, frequency.
- When a case variant is selected, a diagram is shown representing the **average** of all cases in the case variant.
- Updated code in LogStatsAnalyser.getCaseVariantGroupMap() so that the case variant ids remain the same in both case inspector and case variant inspector.


Apromore-Custom-Plugins/Process-Discoverer-Logic/src/main/java/org/apromore/processdiscoverer/ProcessDiscoverer.java
```diff
@@ -125,6 +130,13 @@ public Abstraction generateTraceAbstraction(@NonNull AttributeLog log, @NonNull
        
         return new TraceAbstraction(attTrace, log, params);
     }
+
+    // This method is one-off to view a trace variant
+    public Abstraction generateTraceVariantAbstraction(@NonNull AttributeLog log, @NonNull List<String> traceIDs, @NonNull AbstractionParams params) {
```
brucenguyen1: Please add javadocs comments for this method, the traceIDs parameter needs to be explained: it is supposed to be the list of traceIDs of the same variant? What is the result of this set of traceIDs has some traces of different variant, any error?

Apromore-Custom-Plugins/Process-Discoverer-Logic/src/main/java/org/apromore/processdiscoverer/abstraction/TraceVariantAbstraction.java
```diff
@@ -0,0 +1,63 @@
...
+import org.apromore.processdiscoverer.bpmn.TraceVariantBPMNDiagram;
+import org.apromore.processmining.models.graphbased.directed.bpmn.BPMNDiagram;
+import org.apromore.processmining.models.graphbased.directed.bpmn.BPMNEdge;
+import org.apromore.processmining.models.graphbased.directed.bpmn.BPMNNode;
+
+import java.util.List;
+
+public class TraceVariantAbstraction  extends AbstractAbstraction {
+    public TraceVariantAbstraction(List<AttributeTrace> attTraces, AttributeLog log, AbstractionParams params) {
+        super(log, params);
```
brucenguyen1: Same comment as above, the attTraces list parameter here is supposed to contain only traces of the same variant, i.e. they must have the same sequence of activity, e.g. they must be all "abc", not "acb", what happens if this passing list contains both "abc" and "acb" traces? Just need to document.

Apromore-Custom-Plugins/Process-Discoverer-Logic/src/main/java/org/apromore/processdiscoverer/bpmn/TraceVariantBPMNDiagram.java
```diff
@@ -0,0 +1,91 @@
...
+import java.util.stream.IntStream;
+
+public class TraceVariantBPMNDiagram extends SimpleBPMNDiagram {
+    private MutableObjectLongMap<BPMNNode> nodeDurationMap = ObjectLongMaps.mutable.empty();
+    private MutableObjectLongMap<BPMNEdge<BPMNNode, BPMNNode>> arcDurationMap = ObjectLongMaps.mutable.empty();
+    private BPMNNode startNode;
+    private BPMNNode endNode;
+
+    public TraceVariantBPMNDiagram(List<AttributeTrace> attTraces, AttributeLog log) {
+        super(log);
```
brucenguyen1: Same comment above on the attTraces parameter, it needs some explanation for clarity.

Apromore-Custom-Plugins/Process-Discoverer-Logic/src/main/java/org/apromore/processdiscoverer/ProcessDiscoverer.java
brucenguyen1: We can also check non empty for traceIDs and throw exception because it's no use if it's empty.

Apromore-Custom-Plugins/Process-Discoverer-Logic/src/main/java/org/apromore/processdiscoverer/bpmn/TraceVariantBPMNDiagram.java
```diff
@@ -0,0 +1,91 @@
...
+
+        //Get average durationTrace
+        double[] avgDurationTraces = IntStream.range(0, numDurTraces).boxed()
+                .mapToDouble(i -> IntStream.range(0, attTraces.size())
+                        .mapToDouble(attIndex -> attTraces.get(attIndex).getDurationTrace().get(i))
+                        .sum() / Double.valueOf(attTraces.size()))
+                .toArray();
+
+        Map<Integer, BPMNNode> createdNodes = new HashMap<>();
+        for(int i=1; i<valueTrace.size(); i++) {
```
brucenguyen1: This code is using the 1st trace in attTraces as the main one to build the diagram and measures. If one trace in attTraces is different from the 1st one, e.g. it's shorter, error can happen at line 31.
To be safe and solve the above question throughout, we should add a input check to make sure all traces in attTraces have the same valueTrace, i.e. they are of the same variant.

Apromore-Custom-Plugins/Process-Discoverer-Logic/src/main/java/org/apromore/processdiscoverer/ProcessDiscoverer.java
brucenguyen1: As discussed below, suggest to add an input check early here to check all traces in the traceIDs should be of the same variant, and these IDs must be valid as well, i.e. they can be found in the AttributeLog.

Apromore-Custom-Plugins/Process-Discoverer-Logic/src/test/java/org/apromore/processdiscoverer/ProcessDiscovererTest.java
```diff
@@ -985,6 +1017,73 @@ public void testTraceAbstraction_NodeDuration() {
         }
         
     }
+
+    @Test
+    public void testTraceVariantAbstraction_Sepsis() {
+        try {
+            String traceIDs[] = {"B"};
+            Abstraction abs = discoverTraceVariantAbstraction(read_Sepsis(), new ArrayList<>(Arrays.asList(traceIDs)));
```
brucenguyen1, line 1025: Same comment as above, it would be great if we can add a unit test to check if the input check works properly if passing invalid traceIDs param, e.g. traceIDs with non-existent id, or traceIDs with different trace variants.

Apromore-Custom-Plugins/Process-Discoverer-Portal-Plugin/src/main/java/org/apromore/plugin/portal/processdiscoverer/PDAnalyst.java
```diff
@@ -469,6 +505,43 @@ public boolean hasSufficientDurationVariant(String attribute, String inDegree, S
         return listResult;
     }
 
+    /**
+     * Create a map with the averages of attributes for the activity at the given index of a case variant.
+     * The sequence of activities should be the same for each case of the same case variant.
+     * @param caseVariantID the id of the case variant.
+     * @param index the index of the activity in the cases of this case variant.
+     * @return
```
brucenguyen1: It would be helpful to add an explanation for the return type: what is the key, what is the value.

Apromore-Custom-Plugins/Process-Discoverer-Portal-Plugin/src/main/java/org/apromore/plugin/portal/processdiscoverer/PDAnalyst.java
```diff
@@ -234,6 +242,16 @@ public OutputData discoverTrace(String traceID, UserOptionsData userOptions) thr
         return new OutputData(traceAbs, traceVisualization);
     }
 
+    public OutputData discoverTraceVariant(int traceVariantID, UserOptionsData userOptions) throws Exception {
+        List<ATrace> traces = caseVariantGroupMap.get(traceVariantID);
+        List<String> traceIDs = traces.stream().map(ATrace::getCaseId).collect(Collectors.toList());
+
+        AbstractionParams params = genAbstractionParamsForTrace(userOptions);
+        Abstraction traceAbs = processDiscoverer.generateTraceVariantAbstraction(attLog, traceIDs, params);
```
brucenguyen1: To be safe, it should check whether traceIDs is non-empty and throw exception. This can be checked here or the exception can be thrown from processDiscoverer.generateTraceVariantAbstraction(), better to check here asap right below line 246.

Apromore-Custom-Plugins/Process-Discoverer-Portal-Plugin/src/test/java/org/apromore/plugin/portal/processdiscoverer/PDAnalystTest.java
```diff
@@ -253,5 +268,19 @@ public void test_Filters() throws Exception {
         assertEquals(4, attLog.getTraces().get(1).getValueTrace().size());
 
     }
+
+    @Test
+    public void test_getActivityAttributeAverageMap() throws Exception {
+        PDAnalyst analyst = createPDAnalyst(readLogWithTwoTraceOneVariant());
```
brucenguyen1: It would be great if we can add two unit tests for getCaseVariantDetails() and getActivityAttributeAverageMap() after applying filter to PDAnalyst.

brucenguyen1 (approved):

PR #1242, opened 2021-08-31
AP-3243

chiichang:
support CustomCalendar function for APMLog


Apromore-Extras/APMLogModule/src/main/java/org/apromore/apmlog/logobjects/AbstractLogImpl.java
```diff
@@ -101,4 +104,8 @@ protected void setActivityInstances(List<ActivityInstance> activityInstances) {
     public void setTimeZone(String timeZone) {
         this.timeZone = timeZone;
     }
+
+    public void setCalendarModel(CalendarModel calendarModel) {
+        this.calendarModel = calendarModel;
+    }
```
brucenguyen1: It's better to add a Lombok's @NonNull constraint, e.g. setCalendarModel(@NonNull CalendarModel calendarModel). It will help the debugging of NullPointerException because it will be thrown right in this set method if it happens rather than waiting until some methods encounter null and throw it.

Apromore-Extras/APMLogModule/src/main/java/org/apromore/apmlog/filter/typefilters/PathFilter.java
```diff
@@ -299,14 +300,18 @@ private static boolean haveSameAttributeValue(ActivityInstance act1, ActivityIns
...
         double duration = 0;
 
         if (act2 == null) {
             duration = act1.getDuration();
         } else {
             long st = act1.getStartTime();
             long et = act2.getEndTime();
-            duration = et > st ? et - st : 0;
+            duration = calendarModel != null ?
+                    Long.valueOf(calendarModel.getDuration(new long[]{st}, new long[]{et})[0]).doubleValue() :
```
brucenguyen1: Can we add a utility method, e.g. CalendarDuration.getDuration(CalendarModel calendarModel), add the above code to that method, and then use this method everywhere? It is to the above code duplicated in many places, and some places may forget to do it correctly leading to errors.

brucenguyen1: Can we use the simpler API provided in CalendarModel (below) rather than the above method with conversion to long array? 
- DurationModel getDuration(Long starDateTimeUnixTs, Long endDateTimeunixTs)
Can use DurationModel.getDuration.getNano() or DurationModel.getDuration.get(ChronoUnit.MILLIS).

Apromore-Extras/APMLogModule/src/main/java/org/apromore/apmlog/logobjects/AbstractLogImpl.java
brucenguyen1: Another comment is it's safer to initialize CalendarModel in the APMLog constructor rather than via setCalendarModel method whiich is separated from the constructor and can be forgotten to call. Without a calendar model, all duration computation for the log won't work, that's the reason why it's safer to set calendar model in APMLog constructor  and check for non-null. With it initialized in the constructor, Once APMLog has been created, it is always safe to use its calendar model, no need to check for calendarModel != null condition in many places in the code.

raboczi: If possible, marking `calendarModel` as `final` would allow the compiler to confirm that it's not changed anywhere other than in the constructor.

chiichang: Adding CalendarModel to constructor or making it final are not applicable.

chiichang: making calendarModel final is not applicable

brucenguyen1: That's ok. I remembered now APMLog is cached when it's not associated with any calendar.

Apromore-Extras/APMLogModule/src/main/java/org/apromore/apmlog/filter/typefilters/PathFilter.java
chiichang: DurationModel.getDuration.getNano() not applicable
DurationModel.getDuration.get(ChronoUnit.MILLIS) caused error

brucenguyen1: Pls use Duration.toMillis: https://docs.oracle.com/javase/8/docs/api/java/time/Duration.html#toMillis--

brucenguyen1 (approved):

PR #1244, opened 2021-08-31
AP-4620/error-parsing-date-with-MMM-in-en_AU

Janite:
Added locale to dateTimeFormatter in DateUtils.

If the locale is en_AU, months in MMM format are Aug. instead of Aug (note the ".")


franklma (approved):
Tested with various locales

PR #1246, opened 2021-09-02
AP-4497 Simulator i18n

dreamind:


Janite (approved):
Built and run it in Japanese. Looks good!

PR #1248, opened 2021-09-02
AP-4403: Refactor ZK theming

raboczi:
This PR removes the custom Chiaroscuro Compact ZK theme in favor of the stock Iceblue Compact instead.  Because our overriding CSS stylesheets remain, this _ought_ to cause no change.  The Gradle configuration retains the current ZK version but only sets it in one place.  The motivation for all this [yak-shaving](https://en.wiktionary.org/wiki/yak_shaving) is to ease the ZK9 migration.  This PR is a prerequisite for its ApromoreEE counterpart https://github.com/apromore/ApromoreEE/pull/610.

While the "Files changed" count looks alarming, it's mostly deleted files comprising the source code of Chiaroscuro Compact.


raboczi:
Superseded by https://github.com/apromore/ApromoreCore/pull/1260

PR #1251, opened 2021-09-06
AP-4642 Fix left-over bimp errors

dreamind:


Janite (approved):
Builds and runs. The gateway error no longer occurs, as expected.

PR #1254, opened 2021-09-06
AP-4646

chiichang:
- fixed the code which was not updated for CustomCalendar
- added corresponding UnitTest

### DIFF
--- a/Apromore-Extras/APMLogModule/src/test/java/org/apromore/apmlog/customcalendartests/CusCalArcDurFIlterTest.java
+++ b/Apromore-Extras/APMLogModule/src/test/java/org/apromore/apmlog/customcalendartests/CusCalArcDurFIlterTest.java
@@ -1,0 +1,97 @@
+/**
+ * #%L
+ * This file is part of "Apromore Enterprise Edition".
+ * %%
+ * Copyright (C) 2019 - 2021 Apromore Pty Ltd. All Rights Reserved.
+ * %%
+ * NOTICE:  All information contained herein is, and remains the
+ * property of Apromore Pty Ltd and its suppliers, if any.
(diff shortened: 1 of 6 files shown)


Apromore-Extras/APMLogModule/src/test/java/org/apromore/apmlog/customcalendartests/CusCalArcDurFIlterTest.java
```diff
@@ -0,0 +1,97 @@
...
+import org.apromore.calendar.builder.CalendarModelBuilder;
+import org.apromore.calendar.model.CalendarModel;
+import static org.junit.Assert.assertEquals;
+
+import java.util.Arrays;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+
+public class CusCalArcDurFIlterTest {
```
raboczi: Surely this should be `CusCalArcDurFilterTest'?

raboczi (changes requested):
Just a fiddling typo in a class name, otherwise I'm assuming the original tests were wrong by a millisecond and the changes correct that.

Apromore-Extras/APMLogModule/src/test/java/org/apromore/apmlog/customcalendartests/CusCalArcDurFIlterTest.java
raboczi: Now that I look at it more closely (since SonarCloud pointed it out) this probably should either be named something just *Utility.java rather than *Test.java (since the latter implies it's a test suite) or simply be a private method of APMLogUnitTest if it's unlikely to be reused.

raboczi (approved):

PR #1258, opened 2021-09-07
AP-4499 Add calendar UI to PD

dreamind:


dreamind:
Work still needed to update/reload the PD after custom calendar is updated/restored. This will wait for https://github.com/apromore/ApromoreCore/pull/1255

Apromore-Custom-Plugins/Process-Discoverer-Portal-Plugin/src/main/java/org/apromore/plugin/portal/processdiscoverer/PDController.java
```diff
@@ -451,6 +434,10 @@ public void openSharingWindow() {
         }
     }
 
+    public void openCalendar() {
+        ((MainController)portalContext.getMainController()).getBaseListboxController().launchCalendar(sourceLogName, sourceLogId);
+    }
+
```
brucenguyen1, line 467: As discussed, there is a lack of a communication from CustomCalendar to the hosting plugins like PD. The current issue is after a calendar is changed, e.g. from 27/7 to Monday only, all time measures and statistics in PD still show the values with the old calendar, not the new one.

dreamind:
@brucenguyen1 I have updated the PR and just use the ZK Queue to do the sync.

Apromore-Core-Components/Apromore-Portal/src/main/java/org/apromore/portal/dialogController/BaseListboxController.java
```diff
@@ -842,22 +842,20 @@ protected void share() {
...
   }
 
-  protected void launchCalendar(String artifactName, Integer logId) {
+  public void launchCalendar(String artifactName, Integer logId) {
     PortalPlugin calendarPlugin;
-    getMainController().eraseMessage();
-
-   final EventQueue<Event> queue = EventQueues.lookup("org/apromore/service/CALENDAR", true);
-
+    final EventQueue<Event> sessionQueue = EventQueues.lookup("org/apromore/service/CALENDAR", EventQueues.SESSION,true);
```
brucenguyen1: Hard-coded queue name. Should refer to a static constant provided by CalendarEvents (not by CalendarService), refer to a related comment down below.

Apromore-Core-Components/Apromore-Portal/src/main/java/org/apromore/portal/dialogController/BaseListboxController.java
```diff
@@ -842,22 +842,20 @@ protected void share() {
...
+    sessionQueue.subscribe(new EventListener<Event>() {
       @Override
       public void onEvent(Event event) {
         if ("onCalendarPublish".equals(event.getName())) {
           Long data = (Long) event.getData();
           getMainController().getEventLogService().updateCalendarForLog(logId, data);
+          sessionQueue.publish(new Event("onCalendarChanged", null, logId));
         }
-        queue.unsubscribe(this);
+        sessionQueue.unsubscribe(this);
```
brucenguyen1: This listener is causing issue of updating calendar set for one log to another log (open multiple PD windows for different logs, open the calendar for one log A of these from the Portal, then change the calendar in one PD window for log A, note that other PD windows of other logs also change to the same calendar). This is because it only relies on "onCalendarPublish" event name to do its action which is not enough, it needs to know which logId to do.

Suggest a change: it is unnecessary to put the action of updating calendar for log within this session event listener. This action can be simply put inside the onClickPublishBtn() of the Calendars controller. In this way, the calendar for each logId opened in the CalendarUI is automatically isolated.
No need to implement it via event queue which adds more complexity to manage. Btw, this BaseListboxController has the role of opening the Calendars UI and that's all it should do.

Apromore-Custom-Plugins/Calendar-Portal/src/main/java/org/apromore/plugin/portal/calendar/controllers/Calendar.java
```diff
@@ -522,8 +522,8 @@ public void onClickApplyBtn() {
     @Listen("onClick = #cancelBtn")
     public void onClickCancelBtn() {
         if (isNew) {
-            EventQueue<Event> calendarEventQueue = EventQueues.lookup(CalendarService.EVENT_TOPIC, EventQueues.DESKTOP,true);
-            calendarEventQueue.publish(new Event(CalendarEvents.ON_CALENDAR_ABANDON, null, calendarId));
```
brucenguyen1, line 526: The UI queue name should not be put inside CalendarService which is only backend service for reading/writing calendar in the database. These pure-UI queue names are best put inside the current CalendarEvents class in the CalendarPortal as constants, avoid another dependency on the CalendarService for this naming.

Apromore-Custom-Plugins/Calendar-Portal/src/main/java/org/apromore/plugin/portal/calendar/controllers/Calendars.java
```diff
@@ -200,7 +193,7 @@ public void onClickPublishBtn() {
         String msg = getLabels().getString("success_apply_message");
         String infoText = String.format(msg, logName);
         Notification.info(infoText);
-        calendarEventQueue.publish(new Event(CalendarEvents.ON_CALENDAR_PUBLISH, null,
+        sessionCalendarEventQueue.publish(new Event(CalendarEvents.ON_CALENDAR_PUBLISH, null,
                 ((CalendarModel) calendarListModel.getSelection().iterator().next()).getId()));
         getSelf().detach();
```
brucenguyen1: Using sessionCalendarEventQueue here to send ON_CALENDAR_PUBLISH event is unnecessary. As suggested above, just move the action code inside BaseListboxController to right here.

Using sessionCalendarEventQueue to send onCalendarChanged event to other plugin instances is fine. It should be put here.

Apromore-Custom-Plugins/Calendar-Portal/src/main/java/org/apromore/plugin/portal/calendar/controllers/Calendars.java
```diff
@@ -242,7 +235,7 @@ public void onSelectCalendarItem() {
 
     @Listen("onClick = #restoreBtn")
     public void onClickRestoreBtn() {
-        calendarEventQueue.publish(new Event(CalendarEvents.ON_CALENDAR_PUBLISH, null,null));
+        sessionCalendarEventQueue.publish(new Event(CalendarEvents.ON_CALENDAR_PUBLISH, null,null));
         getSelf().detach();
```
brucenguyen1: As suggested above, make the action code when calendar is published as a common method with parameters in the Calendars UI and use it here.

Use sessionCalendarEventQueue to publish the onCalendarChanged event here for plugins to know.

Apromore-Core-Components/Apromore-Portal/src/main/java/org/apromore/portal/dialogController/BaseListboxController.java
brucenguyen1: The use of onCalendarChanged event to communicate with plugin instances is fine, but it can be simply be initiated from the onClickPublishBtn in CalendarUI, not in BaseListboxController. To be consistent, "onCalendarChanged" should be made constant in CalendarEvents.

Apromore-Custom-Plugins/Process-Discoverer-Portal-Plugin/src/main/java/org/apromore/plugin/portal/processdiscoverer/PDController.java
```diff
@@ -451,6 +438,30 @@ public void openSharingWindow() {
...
+                            }
+                        );
+                    }
+                    sessionQueue.unsubscribe(this);
+                }
+            }
+        });
+        ((MainController)portalContext.getMainController()).getBaseListboxController().launchCalendar(sourceLogName, sourceLogId);
+    }
+
```
brucenguyen1: - Queue name "org/apromore/service/CALENDAR" should use constant to avoid duplicate code, refer to above comment.
- The use of "sessionQueue.unsubscribe(this)" is unnecessary at this stage. If click Open Calendar in PD and then click Cancel to close, this event listener is still registered already, and so when changing calendar in another PD window, this PD window will also ask to refresh the page. It doesn't ask to refresh when the Calendar window is closed NOT via the Cancel button (so it is inconsistent). At this stage, we can accept this behavior and leave the listener still subscribed, but it seems unharmful as it doesn't affect other PD window (separate ZK desktop). It takes time to implement how to clean it up when a ZK desktop is destroyed; for example, PD has a clean up process via PDAnalyst (cleanUp method) via ZK Desktop lifecycle hook.
- Line 442 to 461 should be put inside the initialize code of this Controller, not efficient to put the queue listener in openCalendar because it will be executed every time opening the calendar, and subscribing new listener object/function each time.

dreamind:
@brucenguyen1 Further updated to address your comments.

brucenguyen1 (approved):

PR #1263, opened 2021-09-09
AP-4381-4383 Adjust width and add flag to lang chooser

dreamind:


raboczi (approved):
In Safari and Firefox the locale display is slightly clipped while the drop-down isn't dropped down.  There's nothing here I'm unwilling to see merged, however.

dreamind:
> In Safari and Firefox the locale display is slightly clipped while the drop-down isn't dropped down. There's nothing here I'm unwilling to see merged, however.

This will be rectified as part of ZK 9.6 fixes

PR #1269, opened 2021-09-13
AP-4630 Configure "Perspective" tag in Log Importer

franklma:
Add "perspective" tag in Log importer, and apply to both CSV, Excel and Parquet files.

`org.apromore.service.EventLogService#getPerspectiveTagByLog` is exposed and can be used to get perspective tags for one specified log.

Note: The "Perspective" tag only works for newly imported logs currently. Data migration is needed for existing Logs, and it's still under development.


Apromore-Core-Components/Apromore-Manager/src/main/java/org/apromore/service/EventLogService.java
```diff
@@ -152,4 +153,12 @@ boolean saveFileToVolume(String filename, String prefix,
...
 
   ConfigBean getConfigBean();
+
+
+  /**
+   * Find perspective tag that are linked to specified Log
+   * @param logId Log Id
+   * @return Perspective tags
+   */
+  List<String> getPerspectiveTagByLog(Integer logId) throws JsonProcessingException;
```
brucenguyen1: Instead of JsonProcessingException which is an implementation detail (Jackson library), can we throw our domain-specific exception. At the client layer (Portal plugin), it only knows about domain excetion. For example, in this case, it makes sense to throw a UsermetadataException to represent any abnormal behavior in processing Usermetadata related: invalid dashboard json, invalid perspective tags, invalid filter json, etc.

Apromore-Core-Components/Apromore-Manager/src/main/java/org/apromore/service/impl/EventLogServiceImpl.java
```diff
@@ -565,4 +573,30 @@ public boolean saveFileToVolume(String filename, String prefix, ByteArrayOutputS
...
+
+		String jsonString = usermetadataSet.iterator().next().getContent();
+		ObjectMapper objectMapper = new ObjectMapper();
+
+		try {
+			perspectiveList = objectMapper.readValue(jsonString, new TypeReference<>() {
+			});
+		} catch (JsonProcessingException e) {
+			LOGGER.error("Could not deserialize JSON content from given JSON content String: " + jsonString, e);
+		}
```
brucenguyen1: This exception should be handled properly to indicate something wrong in the data for timely repair. As commented above, suggest to throw a domain-specific exception instead of an implementation-specific one: UsermetadataException. The UI layer will need to catch and display the error message properly, e.g. "Internal error when processing log metadata for perspectives".

Apromore-Core-Components/Apromore-Manager/src/test/java/org/apromore/test/service/impl/EventLogServiceImplTest.java
```diff
@@ -261,4 +294,86 @@ private XLog readXESFile(String fullFilePath) throws Exception {
...
+    // Mock call
+    List<String> perspectives = eventLogService.getPerspectiveTagByLog(logId);
+
+    List<String> result = new ArrayList<>();
+
+    assertEquals(perspectives, result);
+    exceptionRule.expectMessage("Could not deserialize JSON content from given JSON content String: " + invalidJSONString);
+    exceptionRule.expect(JsonProcessingException.class);
+  }
+
```
brucenguyen1: The current test is always a success with any exception because eventLogService.getPerspectiveTagByLog actually doesn't throw any exception. As mentioned above, a domain exception should be used instead and the implementation method should throw the domain exception instead of silent logging. Testing the error message is not as important as testing the thrown Exception class.

exceptionRule.expect() should be before the call eventLogService.getPerspectiveTagByLog().

Can use "@Test(expected = JsonProcessingException.class)" for short.

Apromore-Core-Components/Apromore-Manager/src/test/java/org/apromore/test/service/impl/EventLogServiceImplTest.java
```diff
@@ -99,6 +109,16 @@
   private CustomCalendarService calendarService;
   private AuthorizationService authorizationService;
 
+  public ExpectedException exceptionRule = ExpectedException.none();
+
```
brucenguyen1: Should add @Rule annotation.

Apromore-Custom-Plugins/Log-Importer-Portal/src/main/java/org/apromore/plugin/portal/logimporter/LogImporterController.java
```diff
@@ -398,23 +406,48 @@ protected LogModel getLogModel() throws Exception {
...
+        List<Integer> perspectivePos = logMetaData.getPerspectivePos();
+        List<String> headers = logMetaData.getHeader();
+
+        if (logMetaData.getActivityPos() != HEADER_ABSENT) {
+            result.add(XConceptExtension.KEY_NAME);
+        }
+
+        if (logMetaData.getResourcePos() != HEADER_ABSENT) {
+            result.add(XOrganizationalExtension.KEY_RESOURCE);
+        }
```
brucenguyen1: What happens if the Activity column has actually a different attribute name in the log file from XConceptExtension.KEY_NAME ("concept:name")? For example, if the CSV file being imported doesn't have any attribute with name = "concept:name" but it has an attribute called "action" with the same meaning as "concept:name". If the above code is executed, the list of perspectives will contain "concept:name" rather than "action" which is wrong. Later, when a plugin retrieves this perspective, it won't be able to find any attribute with name = "concept:name" because the log has only attribute with name = "action.

The same for XOrganizationalExtension.KEY_RESOURCE.

brucenguyen1:
As discussed, can we add unit tests for the Perspective data in LogMetaData? These are tests for LogImporterLogic.
Suggest we do the following:
- Move getPerspectiveFromLogMetadata() from LogImporterController to LogMetaData as it should be (rename: getPerspectiveAttributes or getPerspectives)
- Add simple LogMetaData objects for testing along with simple logs for different scenarios, e.g. default concept:name and org;resource prespectives, no default perspectives with a tagged activity attribute, no default/no activity attributes with custom event attribute as perspective.
- Add tests for getPerspectiveFromLogMetadata
- Add tests for LogProcessor.processLog regarding the Perspective data.

Apromore-Custom-Plugins/Log-Importer-Portal/src/main/java/org/apromore/plugin/portal/logimporter/LogImporterController.java
frank-ma-leonardo: > As discussed, can we add unit tests for the Perspective data in LogMetaData? These are tests for LogImporterLogic.
> Suggest we do the following:
> 
> * Move getPerspectiveFromLogMetadata() from LogImporterController to LogMetaData as it should be (rename: getPerspectiveAttributes or getPerspectives)
> * Add simple LogMetaData objects for testing along with simple logs for different scenarios, e.g. default concept:name and org;resource prespectives, no default perspectives with a tagged activity attribute, no default/no activity attributes with custom event attribute as perspective.
> * Add tests for getPerspectiveFromLogMetadata
> * Add tests for LogProcessor.processLog regarding the Perspective data.

Unit tests added as per discussion.

franklma:
> As discussed, can we add unit tests for the Perspective data in LogMetaData? These are tests for LogImporterLogic.
> Suggest we do the following:
> 
> * Move getPerspectiveFromLogMetadata() from LogImporterController to LogMetaData as it should be (rename: getPerspectiveAttributes or getPerspectives)
> * Add simple LogMetaData objects for testing along with simple logs for different scenarios, e.g. default concept:name and org;resource prespectives, no default perspectives with a tagged activity attribute, no default/no activity attributes with custom event attribute as perspective.
> * Add tests for getPerspectiveFromLogMetadata
> * Add tests for LogProcessor.processLog regarding the Perspective data.

Unit tests added as per discussion.

Apromore-Custom-Plugins/Log-Importer-Portal/src/main/java/org/apromore/plugin/portal/logimporter/LogImporterController.java
franklma: The conversion from attribute name to standard XES attribute identifier ("concept:name", "org:resource") is already handled inside `LogProcessor`. So no action is required here as per discussion.

brucenguyen1 (approved):

PR #1278, opened 2021-09-15
AP-4747: Enable password salting

raboczi:
Password management currently has the following significant issues:
* Despite storing a `password_salt` field in the database, the existing login code for locally-stored accounts has never actually salted its password hashes.
* Password hashes are generated using the long-deprecated MD5 algorithm.

This PR addresses these issues by adding a new `membership.password_hashing_algorithm` field in the database which indicates which hashing algorithm to use.  The special value `MD5-UNSALTED` preserves the current behavior.  The likely value to migrate to in the near future is `SHA-256`.  The following feature flags allow migration:
* `passwordHashingAlgorithm` our preferred hashing algorithm used when creating or changing passwords
* `saltLength` number of alphanumeric character in newly-generated salts
* `upgradePasswords` will cause successful logins to user accounts that aren't using `passwordHashingAlgorithm` or which have less than `saltLength` salt to transparently rehash the unchanged password with a new randomized salt
* `allowedPasswordHashingAlgorithms` is a list of the hashing algorithms which are permitted for authentication.  This enables old hashing algorithms to eventually be retired.
 
Currently `passwordHashingAlgorithm` is `MD5-UNSALTED` and `upgradePasswords` is false; this preserves the existing behavior.

The intention in future is for `passwordHashingAlgorithm` to be set to `SHA-256` and `upgradePasswords` to be set to `true` to incrementally improve existing systems.  Eventually `allowedPasswordHashingAlgorithms` can eliminate `MD5-UNSALTED` as an accepted value.

The password checking logic, previously in the persistence layer (UserRepositoryCustom), has been relocated to SecurityService and the Spring Security UsernamePasswordAuthenticationProvider.

There are a number of peripheral changes related to unit tests.  Classes in org.apromore.test.service.impl were moved to org.apromore.service.impl because the tests are intended to have package access to the classes they're testing.  Various minor tweaks to address SonarCloud issues were made in pre-existing Manager tests.


fsmicdev:
Overall, I believe I have got a pretty decent grasp of what you have done here (proposed in first draft of PR), but this will cause a problem witth Keycloak.  What I would need, from a Keycloak perspective, is another column (is_salted [boolean]) in the Membership table which says if a user's password is actually salted.  This is particularly important for the Keycloak password reset use case, so I know whether to store the user's reset password entered as a MD5 hash inculding the pertinent salt.  With this new boolean (is_salted) column, I believe I could refactor the KC user password login and reset password use case related code, so both password authentication and reset password Keycloak use cases works.  Please let me know if you need more information.

fsmicdev (changes requested):
A change is required for Keycloak interoperability: please see my other more detailed comment - essentially a new boolean column in the Membership table is required [ is_salted ], which you will need to write to (can default to 'false' for backward compatibility).

raboczi:
I named the legacy password hashing scheme `MD5 UNSALTED`, which made everything much more explicit.

A gotcha I discovered is that because the password hashes are stored and compared as Strings, they need to be in a canonical form with no leading zeroes.  This isn't a problem internally since all hashes are generated by the `SecurityUtil.hash` method, but it's something that can go wrong if hashes are generated externally and added to the database directly (e.g. Keycloak, or the command line `md5sum` and `shasum` utilities).

raboczi:
Notes arising from the code walkthrough:
* It's possible to configure `passwordHashingAlgorithm` to be a value that isn't in `allowedPasswordHashingAlgorithms`, which means if `upgradePasswords` is enabled the user will not be able to subsequently log in.
* It's possible that a legacy system will have hashed its pre-existing passwords according to the default local character encoding rather than UTF-8, in which case all such passwords will no longer authenticate after this PR is merged.
* The password field is currently limited to 100 characters.  Some hashing algorithms have 512-bit hashes which would require 128 characters.
* If a password is correct but transparent algorithm/salt upgrading fails for any reason, this currently prevents authentication.  The failure of the upgrade should be logged but shouldn't lock the user out.

fsmicdev:
Hi @raboczi Thanks - it was a good walk-through yesterday.  I think you should address points 2 and 3 from your notes above:

* It's possible that a legacy system will have hashed its pre-existing passwords according to the default local character encoding rather than UTF-8, in which case all such passwords will no longer authenticate after this PR is merged.
* The password field is currently limited to 100 characters. Some hashing algorithms have 512-bit hashes which would require 128 characters.

Once solutions are implemented for these two points (along the lines we discussed yesterday [both, in part, require liquibase changes]), and unit/integration tests written testing these two facets, let's catch-up for a quick final run-through.

raboczi:
I addressed all four of the remaining issues.  I renamed `MD5-UNSALTED` because the embedded space was error-prone and changed its behavior to use the platform encoding rather than UTF-8; this makes it compatible with pre-existing databases.  The liquibase changeset was tweaked to use the new name for `MD5-UNSALTED` and also to bump the `password` field to 128 characters so that 512-bit hashing algorithms now work.

fsmicdev (approved):
This looks fine now.  It can be merged if you want, however as discussed previously there will be required changes in KC sometime next week by me ( merge only if you want to risk potential end-users turning on the feature flag/s for this ).

PR #1287, opened 2021-09-20
Feature/ap 4753

dreamind:


Apromore-Custom-Plugins/Calendar-Portal/src/main/resources/static/calendar/zul/calendars.zul
```diff
@@ -42,28 +42,28 @@
...
-        <label value="Available custom calendars:"/>
+        <label value="${$composer.labels.available_calendars_text}"/>
       </hbox>
       <listbox multiple="false" sclass="ap-calendars-listbox" id="calendarListbox" vflex="1" hflex="1" nonselectableTags="input">
         <listhead>
-          <listheader align="left" label="Used" width="48px"/>
-          <listheader align="left" label="Name" hflex="1" />
-          <listheader align="left" label="Created on" width="120px" />
-          <listheader align="center" label="Actions" width="108px"/><!-- icons -->
+          <listheader align="left" label="${$composer.labels.applied_text}" width="48px"/>
```
Janite: This label has changed from "Used" to "Applied". The width of the listheader is currently too small for the text "Applied". We should either increase the width or change the text to something smaller.

Janite (changes requested):
Built and run the code. The labels are showing. Just have a comment about one changed label in a listheader since its text is longer than the listheader width.

Janite (approved):

PR #1292, opened 2021-09-22
AP-4498 Confirm calendar deletion

dreamind:


nolantellis (changes requested):
As per our convo, ry putting the delete logic in the calendar controller and out of the calendar renderer

dreamind:
> As per our convo, ry putting the delete logic in the calendar controller and out of the calendar renderer

Changes are implemented as agreed.

nolantellis (approved):

PR #1299, opened 2021-09-28
AP-4738 Refactor name input dialog

dreamind:


raboczi (approved):
I have no trouble with this being merged.  My one high-level thought is that perhaps inputDialog.zul could be a little tidier if it applied a controller class rather than had all its event handlers explicitly installed by InputDialog.java.

dreamind:
> I have no trouble with this being merged. My one high-level thought is that perhaps inputDialog.zul could be a little tidier if it applied a controller class rather than had all its event handlers explicitly installed by InputDialog.java.

Will take care of this to refactor some of common portal features to Apromore.zk

PR #1306, opened 2021-10-01
AP-4739: Store BPMN documents on filesystem 

raboczi:
The Process Service currently stores BPMN documents in MySQL using NativeRepository.  This PR allows it to also store them using the Storage Service, sharing the mechanism currently used by the Event Log Service.  It adds a `process_model_version.storage_id` field to the database.  Each process model version either has `native_id` populated if it's using NativeRepository, or `storage_id` if it's using the storage service.  The unused field should be null.

While the `enableStorageServiceForProcessModels` feature flag in the application configuration is false, process models will be stored using NativeRepository, preserving the current behavior.  While it is true, they will be stored at `storage.path` under the prefix "model", i.e. something like `~/.apromore/Event-Log-Repository/model/` by default.  The system should work correctly with a mix of native- and storage-based process models.

This PR is a prerequisite for https://github.com/apromore/ApromoreEE/pull/673, which allows process models to be migrated from the native table to a storage provider.

### DIFF
--- a/Apromore-Boot/src/main/resources/application.properties
+++ b/Apromore-Boot/src/main/resources/application.properties
@@ -16,6 +16,7 @@ enableCalendar=true
 bpmndiffEnable=true
 enableEtl=false
 enableConformanceCheck=false
+enableStorageServiceForProcessModels=false
 
 enablePP=true
 enableFullUserReg=true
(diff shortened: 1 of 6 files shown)


franklma (approved):
Work as expected

PR #1313, opened 2021-10-06
AP-4823 Add last updated/created for log

dreamind:


raboczi (approved):
Since both logs and process models have a createDate field, putting it on the shared summary class seems correct.  Approved.

franklma (approved):

PR #1323, opened 2021-10-08
Pass the private Nexus repo credentials to the GitHub action build

raboczi:


raboczi:
Other PRs are passing without this change, so I don't think this is the root cause.  Core should not require access to the private Nexus repo in any case.

PR #1324, opened 2021-10-08
AP-4836 BPMN model issue: Unable to update an existing model

franklma:
-  Fix issue when existing models can not be updated
-  Fix unit test suit of ProcessService

### DIFF
--- a/Apromore-Core-Components/Apromore-Manager/src/main/java/org/apromore/service/impl/WorkspaceServiceImpl.java
+++ b/Apromore-Core-Components/Apromore-Manager/src/main/java/org/apromore/service/impl/WorkspaceServiceImpl.java
@@ -42,7 +42,6 @@ import java.util.Set;
 import java.util.TreeMap;
 import javax.inject.Inject;
 import org.apache.commons.lang3.StringUtils;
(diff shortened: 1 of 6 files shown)


Apromore-Core-Components/Apromore-Manager/src/main/java/org/apromore/service/impl/WorkspaceServiceImpl.java
```diff
@@ -42,6 +42,7 @@
 import java.util.TreeMap;
 import javax.inject.Inject;
 import org.apache.commons.lang3.StringUtils;
+import org.apache.tomcat.jni.Proc;
```
raboczi: Proc is bizarre (Tomcat-proprietary JNI!?) and doesn't appear to be used.  Can we remove it?

raboczi (approved):

Apromore-Core-Components/Apromore-Manager/src/main/java/org/apromore/service/impl/WorkspaceServiceImpl.java
raboczi: I went ahead and fixed this myself.

PR #1326, opened 2021-10-08
AP-4835: Configurable storage prefix

raboczi:
This PR makes the storage prefix used for logs and for process models configurable.  The relevant configuration parameters are `storage.processModelPrefix` and `storage.logPrefix`.  Note that the migration tool has not yet been updated to transfer artifacts stored under an old prefix to a new one.  However the system *ought* to cope with a mixture of different prefixes being in use.


franklma:
work as expected except it seems the model prefix is not picked up from `storage.processModelPrefix` when import models

raboczi:
> work as expected except it seems the model prefix is not picked up from `storage.processModelPrefix` when import models

Addressed this.  I can only guess some of my original work was undone by the merge, but new process models should now have the correct prefix.

franklma:
Work as expected

franklma (approved):

PR #1327, opened 2021-10-08
AP-3806: Apply predefined perspective attributes in PD

brucenguyen1:
This PR makes changes to PD to using the list of perspective attribute keys predefined for logs. New unit tests added, code clean up was also done where possible.

Changes: as described in the commits below.

Verify: open PD, import some logs, CSV or XES. For CSV: can set different perspectives by choosing Event Attribute (Perspective) from the list. For XES, note if the log may have concept:name with org:resource or not (they both can be perspectives). Open the log after importing in PD, verify the perspective list has the same perspective attributes as chosen.


franklma (approved):
work as expected

PR #1337, opened 2021-10-12
AP-4885: Merge regression (v8.0)

raboczi:
Fixes regression in Merge tool caused by changes to process model storage.  (v8.0 back-port of https://github.com/apromore/ApromoreCore/pull/1336)


franklma (approved):
work as expected

PR #1343, opened 2021-10-13
AP-4886: Similarity Search regression

raboczi:
Fixed regression of Similarity Search caused by changes to process storage.


franklma (approved):
work as expected

PR #1344, opened 2021-10-13
AP-4886: Similarity Search regression (v8.0)

raboczi:
Fixed regression of Similarity Search caused by changes to process storage.


Janite (approved):
Works as expected

franklma (approved):
work as expected

PR #1346, opened 2021-10-13
v8.0/AP-4897/update-version-override-message

Janite:


raboczi (approved):
I have to assume the japanese translation is correct, but otherwise this is straightforward.

PR #1350, opened 2021-10-13
AP-4895: Process model copy-on-write

raboczi:
Previously, if you used the Copy and Paste functions to create a copy of a process model, changes to the 1.0 version of the new model would incorrectly also modify the original model.  Now a new storage file is created when a copied model is modified.


franklma (approved):
work as expected

PR #1370, opened 2021-10-20
AP-4676 Remove unused jquery and force bump packages

dreamind:
- Remove jquery in portal
- Bump path-parse and url-parse
- Include resolutions in package.json


dreamind:
@brucenguyen1 @Janite @raboczi I just use resolutions to force bump any npm transitive dependencies (which usually contains vulnerabilities). Probably we should consider switch to yarn, which handles vulnerabilities better.

raboczi (approved):
This seems tidy and my cursory smoke testing doesn't show any breakage.  LGTM!

PR #1375, opened 2021-10-20
AP-4947: Improve BPMN library in ID handling

brucenguyen1:
This PR fixes an issue in the BPMN library (prom-bpmn and prom-models): when a BPMN diagram is created, all elements (nodes, edges, data objects, etc) are always assigned new random unique IDs from UUID, this is no matter whether it is created from an existing .bpmn file or BPMN string representation (which stores IDs of elements), or created from scratch (interactively or programmatically). This issue makes it unable for the front-end to match its BPMN model with the back-end result because the element IDs in the result are different from the version kept by the front-end.

This change allows a BPMN diagram to either create new random IDs for elements or reuse existing IDs. Random IDs are used in case the BPMN diagram is created from scratch whereas reusing IDs is needed when the diagram is created from reading an existing .bpmn file or a representation passed between components (e.g. sending exported string representation from browser to the server).

The API of this BPMN library is unchanged to the current modules if they create BPMN diagrams from scratch, e.g. SplitMiner, or BPMN Editor. For new feature such as the new log animation, conformance checking and alignment service, a new API method in BPMNDiagram (setNextId method) is added to allow them to pass the existing ID to create an element. These are covered in the unit tests (UnmarshallTest).

See detailed changes in the list of commits. 

In summary:
- Add ID to the constructor of NodeID, EdgeID and Node and Edge classes (prom-models module), this is to allow to pass IDs obtained for different purposes as explained above.
- Add getNextId and setNextId to the BPMNDiagram interface and implementation, allowing it to be able to either auto generate new IDs or reuse existing IDs.
- Update all classes under apromore.processmining.plugins.bpmn package to use the setNextId method above to pass an existing ID to the BPMNDiagram, so that it can use this id when creating an element. The classes under this page represent various tags as seen in .bpmn files.
- Remove the use of "Original id" attribute added to the AttributeMap in all the classes in the above package. This is to avoid creating two different ways of keeping element IDs leading to consistency issues and nightmare to maintain the code.
- Improve error handling when creating BPMN diagram from importing files (Bpmn , BPMNImportPlugin)
- Add unit tests (UnmarshallTest)
- Others: clean up code, add comments where needed, remove unused classes where possible (BPMNVisualization, XPDLReader), fix a bug in SplitMiner when calling to Gateway constructor improperly (GatewayMap.java), add a Todo to BpmnDataAssociation as it is not handling this element properly.

Note that these two modules (prom-bpmn and prom-models) are legacy code written in research projects. Code quality comments were raised but only addressed where changes were made in this PR. Addressing all the other Sonar comments would require changes in many other places which is unsafe to do in this PR without unit tests to cover the related functions.


raboczi (approved):
I'm satisfied this is good enough to merge.

PR #1387, opened 2021-10-21
AP-4947: fix bug - bpmn.io doesn't support UUID format for element IDs

brucenguyen1:
This PR is to fix a bug since bpmn.io doesn't support UUID format in a recent change, as shown below.

image

Meanwhile, this PR also uses a short IDs for space saving and easier to read when working with BPMN, the new IDs also have an alphabet prefix to add more semantics.

Changes:
- Add LocalIDGenerator: used to create local and short IDs with a prefix, for use within a BPMN diagram
- BPMNDiagram: getNextId to accept a prefix.
- Update BPMNNode, BPMNEdge to use the above.
- Update UnmarshallTest.

Verify: verify the BPMN Diagram generated from PD can be open in BPMN Editor, and generated BPMN diagrams can be merged in the Merge plugin.


Apromore-ProcessMining-Collection/prom-bpmn/src/main/java/org/apromore/processmining/models/graphbased/directed/bpmn/BPMNDiagram.java
```diff
@@ -59,9 +59,10 @@
 	/**
 	 * Generate the next ID used in creating diagram elements
 	 * This method will return any existing IDs added via {@link #setNextId} or it will return a random ID.
+	 * @param idPrefix: prefix to provide some semantics for the ID
 	 * @return: ID used for assigning to new element.
```
raboczi: No colons in these Javadoc annotations

raboczi (approved):
That all seems reasonable.  I have a minor quibble about Javadoc syntax, but otherwise it's safe to merge.

PR #1404, opened 2021-10-26
AP-5021/enable-conformance-checker

Janite:
- Enable conformance checker
- Added alignmentClient.uri to application.properties. This property must be set to use the alignment client in conformance checker.


raboczi (approved):
Allows the server to start again; conformance checker appears in the menu and fails moderately gracefully in the absence of the trace alignment service.

PR #1409, opened 2021-10-28
AP-4900 Support reverse selection for calendar

dreamind:


Apromore-Custom-Plugins/Calendar-Portal/src/main/resources/static/calendar/js/index.js
```diff
@@ -427,7 +438,8 @@ zk.afterMount(function() {
                 currentDOW = get(cell, 'dow');
                 startHour = get(cell, 'hour');
                 startMin = get(cell, 'min');
-                startSelected = isSelected(currentDOW, startHour, startMin);
+                // startSelected = isSelected(currentDOW, startHour, startMin);
```
Janite, line 441: Do we need to keep the commented out code?

Janite (approved):
Works as expected

Apromore-Custom-Plugins/Calendar-Portal/src/main/resources/static/calendar/js/index.js
dreamind: Will remove this after testing, just in case there is a change of mind :)

PR #1410, opened 2021-10-29
AP-5039 Fix process detail selection

dreamind:


Janite (commented):
When I select multiple using Ctrl/Command, the behaviour is as expected. But when I select multiple using the select all button, the details of the previous selected model are still displayed.

Janite (approved):

PR #1412, opened 2021-10-30
AP-4997 Fix missing default label key

dreamind:


Apromore-Custom-Plugins/Process-Discoverer-Portal-Plugin/src/main/java/org/apromore/plugin/portal/processdiscoverer/PDController.java
```diff
@@ -174,15 +174,21 @@ public ResourceBundle getLabels() {
...
     }
 
     public String getLabel(String key, String defaultVal) {
-        String label = getLabels().getString(key);
+        String label;
+        try {
+            label = getLabels().getString(key);
+        } catch (MissingResourceException e) {
+            label = null;
+        }
```
brucenguyen1: Can we replace the method code with:
try {
    return getLabels().getString(key);
}
catch (Exception) {
    return defaultVal;
}

Apromore-Custom-Plugins/Process-Discoverer-Portal-Plugin/src/main/java/org/apromore/plugin/portal/processdiscoverer/PDController.java
```diff
@@ -174,15 +174,21 @@ public ResourceBundle getLabels() {
...
 
     public String getLabel(String key) {
-        String label = getLabels().getString(key);
+        String label;
+        label = getLabel(key, null);
         if (label == null) {
-            label = "";
+            label = key;
         }
         return label;
```
brucenguyen1: Can we replace the above code with one line: return getLabel(key, key)?

brucenguyen1 (approved):

PR #1415, opened 2021-11-02
Gradle build fails on Windows

raboczi:
This PR duplicates a fix from the ApromoreEE repository for failing Gradle builds on Windows.  This is caused by the different separators in the user.home system property.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -96,7 +96,7 @@ subprojects {
   project(':Apromore-Boot') {
 
     bootJar.baseName = "Apromore-Core"
-    ext['user_home'] = System.properties['user.home']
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -96,7 +96,7 @@ subprojects {
   project(':Apromore-Boot') {
 
     bootJar.baseName = "Apromore-Core"
-    ext['user_home'] = System.properties['user.home']
+    ext['user_home'] = System.properties['user.home'].replaceAll("\\\\", "/")
```
raboczi, line 99: Compare to the ApromoreEE equivalent fix on line 166 of https://github.com/apromore/ApromoreEE/blob/development/build.gradle

karthik-balasubramanian (approved):

PR #1442, opened 2021-11-10
AP-5086: Update version number (v8.0)

raboczi:
Bump the version number of the generated bootable .jar to 8.0. This also shows up in the About dialogue box.  There is a counterpart to this for the development branch at https://github.com/apromore/ApromoreCore/pull/1441

Changing the name of the bootable .jar may require a corresponding change in the CD pipeline, so this needs to b confirmed with devops before merging.


raboczi:
Wrong base branch.

PR #1445, opened 2021-11-10
AP-5099: Set maxNumberOfNodes and maxNumberOfArcs to 5000 and 15000 when building graphs

brucenguyen1:
This PR sets these properties to the values as requested. Some small modifications to move these properties from private constants to parameters and configuration data for later externalizing them to a property file.
Details of changes: described in each commit below.
Trial: try with the attached log in the card which contains 5000 nodes.


Apromore-Custom-Plugins/Process-Discoverer-Portal-Plugin/src/main/java/org/apromore/plugin/portal/processdiscoverer/data/ConfigData.java
```diff
@@ -41,19 +42,24 @@
     public static MeasureType DEFAULT_MEASURE_TYPE = MeasureType.FREQUENCY;
     public static MeasureAggregation DEFAULT_MEASURE_AGGREGATE = MeasureAggregation.CASES;
 
-    public static ConfigData DEFAULT = new ConfigData("concept:name", 500);
+    public static ConfigData DEFAULT = new ConfigData("concept:name", 5000, 15000);
```
dreamind, line 45: final here?

dreamind (approved):

PR #1448, opened 2021-11-11
AP-5100: Externalize the two properties maxNumberOfNodes and maxNumberOfArcs to property files

brucenguyen1:
- Create pd.application.properties under PD Portal Plugin with the two properties, set to the values as requested.
- Update ConfigData to read these properties, making it a bean.
- Update PDController to use this bean.
- Verify unit tests are OK.


raboczi (approved):
Seems well-reasoned.

PR #1454, opened 2021-11-12
feature/AP-2541 simulate model from bpmn editor

Janite:
This PR has a pair in EE: https://github.com/apromore/ApromoreEE/pull/820

This PR removes the "toggle simulation parameters" button and adds the "simulate model" button to the toolbar in bpmn editor. The "simulate model" button opens the Simulate Model plugin which was previously only available from the analyze menu in portal.

Please note: since "Simulate model" is an EE feature, the button is disabled when running Core by itself.

Other changes:
- Remove the extra separator between zoom in and zoom out in bpmn editor


Apromore-Frontend/src/processmap/bpmnModelWrapper.js
```diff
@@ -45,7 +45,8 @@ export default class BPMNModelWrapper {
                 window.Apromore.I18N.Undo.undo,
                 window.Apromore.I18N.Undo.redo,
                 window.Apromore.I18N.Share.share,
-                window.Apromore.I18N.SimulationPanel.toggleSimulationDrawer
+                window.Apromore.I18N.SimulationPanel.toggleSimulationDrawer,
+                window.Apromore.I18N.SimulationPanel.simulateModel
```
brucenguyen1: Can we remove simulationPanel.js and references to it as it is now replaced with simulateModel.js?

Apromore-Frontend/src/bpmneditor/plugins/plugins.js
```diff
@@ -2,6 +2,7 @@ import savePlugin from './apromoreSave';
 import exportPlugin from './export';
 import pdfPlugin from './pdf';
 import sharePlugin from './share';
+import simModelPlugin from './simulateModel';
 import simPanelPlugin from './simulationPanel';
```
brucenguyen1: Can we remove simulationPanel.js and its reference as it's not used any more? (same comment below).

Apromore-Core-Components/Apromore-Portal/src/main/java/org/apromore/portal/dialogController/BPMNEditorController.java
```diff
@@ -231,6 +232,39 @@ public void onEvent(final Event event) throws InterruptedException {
...
+
+        if (currentUserAccessType == AccessType.VIEWER) {
+          Notification.error(Labels.getLabel("portal_noPrivilegeSaveEdit_message"));
+          return;
+        }
+
+        //Only allow model to be simulated if it is saved
+        if (isNewProcess || process == null) {
+          Notification.error(Labels.getLabel("portal_saveModelFirst_message"));
+        } else {
```
brucenguyen1: Checking for new process here is not necessary because the simulation can work on unsaved model, we can remove this check without problems. Note that it's different from "onShare" event handling because it makes no sense to share an unsaved model (a share record will be created in the database with reference to the saved model, thus a new model must  be saved first to have a record in the database).

Apromore-Core-Components/Apromore-Portal/src/main/java/org/apromore/portal/dialogController/BPMNEditorController.java
```diff
@@ -198,6 +198,7 @@ public BPMNEditorController() {
         param.put("bpmnioLib", AccessType.VIEWER.equals(currentUserAccessType) ? BPMNIO_VIEWER_JS : BPMNIO_MODELER_JS);
       }
       param.put("viewOnly", AccessType.VIEWER.equals(currentUserAccessType));
+      param.put("disableSimulateModel", mainC.getPortalPluginMap().get(PluginCatalog.PLUGIN_SIMULATE_MODEL) == null);
       Executions.getCurrent().pushArg(param);
```
brucenguyen1: A better name for "disableSimulateModel" is "availableSimulateModelPlugin" which is related to the checked condition. When this param is passed and checked in the .zul file below, it would be if (${arg.availableSimulateModelPlugin})) which is easier to understand why the button needs to be disbled. 

Somebody reading this code later can get the idea and apply the same to other EE plugins to be called from the editor.

Apromore-Core-Components/Apromore-Portal/src/main/java/org/apromore/portal/dialogController/BPMNEditorController.java
```diff
@@ -231,6 +232,39 @@ public void onEvent(final Event event) throws InterruptedException {
...
+
+        //Only allow model to be simulated if it is saved
+        if (isNewProcess || process == null) {
+          Notification.error(Labels.getLabel("portal_saveModelFirst_message"));
+        } else {
+          Map arg = new HashMap<>();
+          arg.put("selectedModel", process);
+          arg.put("modelData", eventToString(event));
+          simulateModelPlugin.setSimpleParams(arg);
+          simulateModelPlugin.execute(portalContext);
```
brucenguyen1: The setSimpleParams looks like not thread-safe and it won't work properly in a multi-user environment. For example, a user calls this plugin while another user calls this plugin concurrently. 

The parameters should be passed as a method param rather than via setting as an instance variable of the plugin, for example: plugin.execute(PortalContext, SimpleParams), this will avoid thread conflict because each thread has a separate stack frame.

As this is an architectural issue, suggest we put a @todo item in the header of BPMNEditorController, e.g. @todo: avoid thread conflict issues when setting instance data for BIMPPlugin, instead it should pass as a method parameter.

Apromore-Core-Components/Apromore-Portal/src/main/java/org/apromore/portal/dialogController/BPMNEditorController.java
Janite: Removed check for new/saved process. Logs created by simulating unsaved models now have a suggested name of "untitled_sim_scenario".

brucenguyen1 (approved):

PR #1461, opened 2021-11-15
AP-5082: Improve Apromore-Calendar to fix bugs

brucenguyen1:
This PR remedies a number of defects in Apromore-Calendar. The main cause is that the previous implementation was heavily based on a procedural ago in calculating duration: it jumps each day, in each day it checks whether the day falls into work day, then it checks if the day falls into holidays, etc. This is very hard to understand and error-prone; actually it has bugs in handling holidays along this calculation path, and other errors below:
- Mistakes in checking if a day overlaps with a holiday or not.
- Ignore the difference in timezone between start and end
- Imprecise in handling micro durations: e.g. duration of 1 second.

This PR replaces the above procedural algo with interval-based implementation based on streams. All work days and holidays are streams of intervals. So, the duration is to include all work day intervals and exclude all holiday intervals, then calculation the intersection with the [start, end] interval. The main code is in Duration getDuration(Instant start, Instant end).

Look for more details in the unit tests: 
DurationCalculationTestWithHolidayUnitTest
DurationCalculationTestWithMicroDurationUnitTest

### DIFF
--- a/Apromore-Calendar/src/main/java/org/apromore/calendar/model/CalendarModel.java
+++ b/Apromore-Calendar/src/main/java/org/apromore/calendar/model/CalendarModel.java
@@ -32,24 +32,29 @@
 
 package org.apromore.calendar.model;
 
-import java.time.DayOfWeek;
-import java.time.Duration;
-import java.time.Instant;
-import java.time.LocalDate;
-import java.time.OffsetDateTime;
-import java.time.OffsetTime;
-import java.time.ZoneId;
-import java.time.ZoneOffset;
-import java.time.ZonedDateTime;
+import lombok.Data;
+import net.time4j.ClockUnit;
+import net.time4j.Moment;
+import net.time4j.range.ChronoInterval;
+import net.time4j.range.IntervalCollection;
+import net.time4j.range.MomentInterval;
+import net.time4j.tz.Timezone;
+import org.apache.commons.collections.map.HashedMap;
+
+import java.time.*;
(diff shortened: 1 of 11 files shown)


Apromore-Calendar/src/main/java/org/apromore/calendar/model/CalendarModel.java
```diff
@@ -169,67 +126,29 @@ public DurationModel getDuration(Long starDateTimeUnixTs, Long endDateTimeunixTs
...
-
-    Duration duration = Duration.between(workDayModel.getAdjustedStartTime(startTime),
-        workDayModel.getAdjustedEndTime(endTime));
-
-    return duration.isNegative() ? Duration.ZERO : duration;
-
+  private List<ChronoInterval<Moment>> getHolidayIntervals() {
+    return holidayIntervals != null
+            ? holidayIntervals
+            : (holidayIntervals = holidays.stream()
```
raboczi: This evaluation mixed with assignment of `holidayIntervals` is tricky enough that I'd request it be changed.

raboczi (changes requested):

Apromore-Calendar/src/main/java/org/apromore/calendar/model/CalendarModel.java
brucenguyen1: Done

raboczi (approved):

PR #1469, opened 2021-11-17
AP-5184: Improve validity check for BPMN models

brucenguyen1:
This PR fixes some wrong model checking messages for AP-5184 and AP-5154. The issue is that the error message is not informative and there's a bug when some code runs on invalid models before the model is checked for validity.

Changes:
- Move model checking out of Replayer.replayXXX and make it the first step in LogAnimationServiceImpl2.animateXX.
- Strengthen the checking conditions in BPMNDiagramHelper.checkModel, add ModelCheckResult class.
- Clean up code where possible: remove LogAnimationServiceImpl as it is unused.
- Unit tests are added to ModelValidityTest
- Many test samples were added for different types of invalid models.


Apromore-Custom-Plugins/Log-Animation-Logic/src/main/java/org/apromore/service/loganimation/replay/BPMNDiagramHelper.java
```diff
@@ -95,150 +85,134 @@
...
+                    }
+                    targets.put((FlowNode) element, targetSet);
+
+                    sourceSet = new HashSet();
+                    for (SequenceFlow sflow : ((FlowNode) element).getIncomingSequenceFlows()) {
+                        sourceSet.add((FlowNode) sflow.getSourceRef());
+                    }
+                    sources.put((FlowNode) element, sourceSet);
+
+                    if (isDecision((FlowNode) element)) {
```
dreamind, line 165: Are these conditions mutually exclusive? If that's the case, perhaps switch would be better?

dreamind (approved):
There are a couple issues from sonar that can be addressed

PR #1478, opened 2021-11-18
Feature/ap 5058 part 2

dreamind:


Apromore-Custom-Plugins/Account-Portal-Plugin/src/main/resources/account.properties
```diff
@@ -0,0 +1,6 @@
+passwordTooShort = New password must be at least 6 characters long.
```
raboczi, line 1: Pass the value 6 as a parameter, using java.text.MessageFormat to localize it.

raboczi (approved):
This is good enough to merge, but I did include a change suggestion.  Consider it, and the issues raised by SonarCloud.

Apromore-Custom-Plugins/Account-Portal-Plugin/src/main/resources/account.properties
dreamind: Will create a separate ticket to pull this from application.properties

PR #1494, opened 2021-11-23
AP-5250: Cleanup of log upload limiting

raboczi:
This PR tidies up how the limiting of event log uploads is configured.

Previously, CSV uploads were configured using the `isDemoCsv` and `csvDemoMaxLineCount` properties, Excel and parquet uploads were configured using the `maxEventCount` property, and the as-yet-unused new code that imports directly to parquet rather than XES didn't have limiting at all.

Now all formats should be limited by `maxEventCount`, which can be left unset to indicate no limit.  The default value of `maxEventCount` is set in `application.properties` as 1000000 events.


franklma (approved):
Work as expected

PR #1505, opened 2021-11-24
AP-5282: Folder timestamps

raboczi:
This PR removes all the `@Temporal(TemporalType.DATE)` annotations from the DAOs in the Apromore-Database component.  This was truncating all stored timestamps for Folders, Memberships, Users and Workspaces to the nearest day.  I'm presuming the original author thought the intention was for the DAOs to return SQL dates (accurate to the day) rather than Java dates (accurate to the second).


franklma (approved):
Works as expected.

PR #1508, opened 2021-11-24
AP-5284: Folder modification timestamp now updates

raboczi:
Previously, folder modification timestamps weren't updated when the folder was updated.


franklma (approved):

franklma:
Works as expected

PR #1527, opened 2021-11-29
AP-5186 Prevent Integrator role portal login

Janite:
This PR prevents users with only the "Integrator" role, or no role from logging into the Apromore portal.

Other changes:
- The PortalKeyCloakSecurity config authorities now contain apromore roles instead of keycloak roles.
- Fix link to the access denied page in PortalSecurityConfig


raboczi (approved):
This is very tidy.  👍

PR #1531, opened 2021-11-30
AP-5351-5352 Remove user role

Janite:
The PR removes the "User" role and migrates all existing users with the "User" role to the "Analyst" role.


Apromore-Core-Components/Apromore-Manager/src/main/java/org/apromore/service/impl/SecurityServiceImpl.java
```diff
@@ -313,16 +312,12 @@ public User createUser(User user) {
     user.getMembership().setDateCreated(user.getDateCreated());
 
     // A new user is in the USER and ANALYST role
```
raboczi: Does this comment need updating?

Apromore-Database/src/main/resources/db/migration/changeLog-Data.yaml
```diff
@@ -210,4 +210,34 @@ databaseChangeLog:
...
+            where: roleid = 2 AND
+              userid IN (
+              SELECT * FROM (
+              SELECT userid
+              FROM `user_role`
+              WHERE roleid=2 OR roleid=4
+              GROUP BY userid
+              HAVING COUNT(roleid) > 1
+              )tblTmp
+              );
```
raboczi: Suggest reducing to:
```
where: roleid = 2 AND userid IN (
    SELECT userid
    FROM `user_role`
    WHERE roleid IN (2, 4)
    GROUP BY userid
    HAVING COUNT(roleid) > 1
);
```

raboczi (changes requested):
I have a change to the Liquibase query I'd like considered.

Apromore-Database/src/main/resources/db/migration/changeLog-Data.yaml
Janite: Made slight change to the suggested query since SQL doesn't allow the target table (user_role) to be referenced directly in the FROM clause for a delete:

```
where: roleid = 2 AND userid IN (
    SELECT userid
    FROM (SELECT * FROM `user_role`) AS user_role_copy
    WHERE roleid IN (2, 4)
    GROUP BY userid
    HAVING COUNT(roleid) > 1
);
```

raboczi (approved):
I'm happy with this.

PR #1533, opened 2021-12-01
Improvement/ap 5299/upload log with log importer

mayankgituni:
This PR implements the methods to get folder id from the path name. And makes the storeMetadataAsJSON in LogImporterController a public method for usage in the ETL.


franklma (approved):
works as expected

PR #1539, opened 2021-12-03
AP-5391: Improve Custom-Calendar performance

brucenguyen1:
The previous implementation based on stream of intervals is easy to understand but it has slower performance  due to the heavy number of intervals and Time4J library handling. This version uses stream of durations with early calculation of duration, and not use Time4J. 

It improves the load test (DurationCalculationUnixTsLoadUnitTest) from 2.75s (previously) to 0.75s.

There was a new unit test added DurationCalculationTestWithSpecialTimezones for testing timezones, but it needs more investigation, temporarily disabled.

### DIFF
--- a/Apromore-Calendar/src/main/java/org/apromore/calendar/model/WorkDayModel.java
+++ b/Apromore-Calendar/src/main/java/org/apromore/calendar/model/WorkDayModel.java
@@ -98,18 +98,8 @@ public class WorkDayModel {
   public Duration getWorkDurationAtDate(OffsetDateTime d, OffsetDateTime start, OffsetDateTime end) {
     OffsetDateTime workStart = startTime.atDate(d.toLocalDate());
     OffsetDateTime workEnd = endTime.atDate(d.toLocalDate());
-    if (workStart.isAfter(start) && workEnd.isBefore(end)) {
-      return this.getDuration();
-    }
-    else if (workStart.isBefore(start) && workEnd.isBefore(end)) {
-      return Duration.between(start, workEnd);
-    }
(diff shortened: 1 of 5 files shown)


Apromore-Calendar/src/main/java/org/apromore/calendar/model/WorkDayModel.java
```diff
@@ -91,26 +87,29 @@
...
+    }
+    else if (workStart.isBefore(start) && workEnd.isBefore(end)) {
+      return Duration.between(start, workEnd);
+    }
+    else if (workStart.isAfter(start) && workEnd.isAfter(end)) {
+      return Duration.between(workStart, end);
+    }
+    else {
+      return Duration.between(start, end);
+    }
```
raboczi: Maybe this could be easier, something like:
```
return Duration.between(workStart.isAfter(start) ? workStart : start,
                        workEnd.isBefore(end)    ? workEnd   : end);
```
Since the entire point of the PR is performance improvement, it'd require profiling to see whether calling `Duration.between` more often has a noticeable impact.

raboczi (commented):
I have a suggestion for a simplification, but you're in a better position to confirm it won't impact performance.  The code smells SonarCloud is flagging seem pretty easy to address, as well.

raboczi (approved):

Apromore-Calendar/src/main/java/org/apromore/calendar/model/WorkDayModel.java
brucenguyen1: Done.

PR #1543, opened 2021-12-06
AP-5416 Refactor PermissionType into an enum

Janite:


raboczi (approved):
That worked out really well.

PR #1548, opened 2021-12-06
AP-5376 Implement Viewer (models) role

Janite:
- Add viewer (models) role to db and give it login permission.
- Open models in view mode if a user has view models permission but not edit models permission


franklma (approved):
Works as expected. 

Technically we should not override access type (ACL based access control) using permission type (Role-based access control), which would increase complexities in the logic and may get messy in scale. I incline to create a separate read-mode entry point for all plugins like BPMN editor.

PR #1552, opened 2021-12-08
AP-5375 Implement viewer role

Janite:
- Change Observer to Viewer and add viewer permissions to db
- Allow users with "View dsicover model" permission to open pd
- Open process discoverer in view mode if the user does not have edit permission


franklma (approved):
LGTM

PR #1561, opened 2021-12-13
AP-5379 Migrate exisiting users to admin, analyst and/or integrator role

Janite:
Migrate existing user roles such that:

- All users with the "integrator" role, keep the "integrator" role.
- Any user with multiple roles, including the "admin" role, only keep their "admin" and "integrator" roles.
- Any user with a role other "admin" and "integrator" should migrate to the "analyst" role.
- Users with no roles should remain with no roles


franklma (approved):
works as expected

PR #1563, opened 2021-12-14
Don't change folder when moving an item via drag&drop. REF: AP-5360

shapon2004:


raboczi (approved):
This looks fine, except that the whitespace/indentation has been a bit scrambled.  Could you set your IDE to always use spaces, never tabs?

shapon2004:
Fixed issues based on comment(on 5360) and issue-5326

PR #1580, opened 2021-12-23
First cut of including processSimulationInfo when exporting BPMN model.

lahiru-maha:
- A new service to retrieve the simulation info
- A new object model for extension elements, and process simulation info
- Including simulation info into the output data at the time of process discovery
- Injecting the simulation info into the bpmn xml at the time of exporting
- adding unit tests
- some minor code cleanup


raboczi (approved):
My only change request was to revert the wildcarded imports.  I understand this will happen in a subsequent PR and it's pretty minor.  I'm jealous of how tidy the XML handling is in this.  ;)

PR #1586, opened 2022-01-05
bugfix/AP-5203 Update APIs to allow for custom column name mapping

Janite:
This PR must be merged alongside its counterpart in EE: https://github.com/apromore/ApromoreEE/pull/954

This PR adds a customHeaderMap paramater to the extractMetaData method which is used to update existing column names with names in the map. It is currently only used when importing parquet files via ETL.


mayankgituni (approved):
Looks good to me!

PR #1600, opened 2022-01-12
Adding default timetable info when exporting bpmn model with simulaton parameters

lahiru-maha:


Apromore-Custom-Plugins/Process-Simulation-Info-Logic/src/main/java/org/apromore/processsimulation/service/SimulationInfoService.java
```diff
@@ -59,23 +72,15 @@
     private static final String XML_START_DEFINITIONS_TAG = "<definitions";
     private static final String XML_START_BPMN_DEFINITIONS_TAG = "<bpmn:definitions";
     private static final String XML_QBP_NAMESPACE = "\n xmlns:qbp=\"http://www.qbp-simulator.com/Schema201212\"\n";
-    private static final String NAN = "NaN";
+    private static final Locale ENGLISH_LOCALE = Locale.ENGLISH;
```
raboczi: This isn't well-named.  It should be something like `DOCUMENT_LOCALE`.

raboczi (approved):
This is good enough to merge, but I had one super-nitpicky comment about a constant name.

PR #1601, opened 2022-01-13
Swapped out the deprecated jcenter repository for Maven Central.

raboczi:
This PR removes the use of JFrog's jcenter artifact repository from the Gradle build process.  See https://blog.gradle.org/jcenter-shutdown for the original notification that jcenter was at risk.  Jcenter failed to be accessible on January 12 2022: https://status.gradle.com/incidents/nv93msj8q658.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -3,7 +3,7 @@ buildscript {
     maven {
       url "https://plugins.gradle.org/m2/"
     }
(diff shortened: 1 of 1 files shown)


lahiru-maha (approved):
Thanks for this!

PR #1604, opened 2022-01-13
AP-5588 Support custom multiple icons for bpmn element

dreamind:


Janite (approved):
Multiple icons can be added as expected.

PR #1605, opened 2022-01-14
Adding the default resource information to the process simulation info

lahiru-maha:


Apromore-Custom-Plugins/Process-Simulation-Info-Logic/src/main/java/org/apromore/processsimulation/service/SimulationInfoService.java
```diff
@@ -107,10 +109,12 @@ public ProcessSimulationInfo deriveSimulationInfo(
 
                 deriveGeneralInfo(builder, logSummary);
 
-                deriveTaskInfo(builder, abstraction);
+                deriveTaskInfo(builder, (AbstractAbstraction) abstraction);
 
                 deriveTimetable(builder);
 
+                deriveResourceInfo(builder, (AbstractAbstraction) abstraction);
```
raboczi: How about adding `final AbstractAbstraction abstractAbstraction = (AbstractAbstraction) abstraction;` right at the beginning of the conditional block where you test that `abstraction instanceof AbstractAbstraction` holds, and avoid having to cast `abstraction` repeatedly thereafter?

lahiru-maha: Done.

raboczi (approved):

PR #1609, opened 2022-01-16
Feature/ap 5580 commenting

dreamind:


Janite (changes requested):
When I add multiple comments on one activity, save, and re-open the model, only one comment appears on the activity instead of the number expected.

I also have a few questions:
1. Comments can not be added when opening the model in view mode. Is this something that will be added as part of a separate PR?
2. The comments should contain the first and last name of the user who left it. Also want to check if that should be included in this PR or part of another.

Apromore-Core-Components/Apromore-Portal/src/main/java/org/apromore/portal/dialogController/BPMNEditorController.java
```diff
@@ -435,7 +436,7 @@ private void pushViewModeParameters(final String bpmnXML) {
     param.put("viewOnly", true);
     param.put("langTag", getLanguageTag());
     param.put("doAutoLayout", "false");
-
+    param.put("username", currentUserType.getUsername());
```
Janite, line 439: currentUserType is null at this point. We need to move the line where currentUserType is set (line 118) before the view link is opened (lines 112 - 116).

Janite (changes requested):

Janite (approved):

PR #1613, opened 2022-01-18
AP-5645: Add support in PD portal plugin to allow other plugins to open it with log filter criteria

brucenguyen1:
Changes:
- Add PDPluginAPI, the openWithFilters method has logFilters param.
- Changes in PDFrequencyPlugin and PDAbstractPlugin to handle openWithFilters 
- Changes in PDController to handle log filtering when opening.


shapon2004 (approved):
Looks fine!

PR #1617, opened 2022-01-19
AP-5410 Title Name Change based single or multiple selection

shapon2004:
when exporting more than one log, can we change the title of the following window from “Download log” to “Download logs”


franklma (approved):
LGTM

PR #1619, opened 2022-01-19
Adding a conditional checkbox to include or exclude simulation parameters at the time of BPMN export

lahiru-maha:


raboczi (approved):

dreamind (approved):

Apromore-Custom-Plugins/Process-Discoverer-Portal-Plugin/src/main/java/org/apromore/plugin/portal/processdiscoverer/eventlisteners/BPMNExportController.java
```diff
@@ -263,52 +270,62 @@ private void save() throws Exception {
         if (parent.getContextData().getLogName() != null) {
             defaultProcessName = parent.getContextData().getLogName().split("\\.")[0];
         }
+
+        String conditionalMessage =
+            simulationInfoService.isFeatureEnabled() ? parent.getLabel("includeSimulationParams_message") : null;
+
```
brucenguyen1: The condition of "simulationInfoService is not enabled" is combined into the handling of whether the user wants to have simulation data in the exported model. Instead, when "simulationInfoService is not enabled", it's better to have a little warning message somewhere saying "The service to extract the simulation data is not enabled, model exported won't have simulation data", rather than showing nothing and let the user go ahead export it (without the simulation data). This behaviour (if happens) should be handled a bit differently rather than combining it into the same way of handling "when the checkbox is not checked".

brucenguyen1 (commented):
Reviewed with comment to make it more explicit and intuitive in handling the condition "when the service flag for SimulationInfoLogic is not enabled". It is implicit and combined atm.

Apromore-Custom-Plugins/Process-Discoverer-Portal-Plugin/src/main/java/org/apromore/plugin/portal/processdiscoverer/eventlisteners/BPMNExportController.java
lahiru-maha: So the idea of the simulationInfoService.isFeatureEnabled() method is to act as the feature toggle. Meaning for some clients this could be purposefuly toggled off. I wonder if in that case displaying a warning message is a good idea, and might cause confusion? i.e. warning for a feature they were never meant to have in the first place. Since the simulator is only meant for enterprise clients, it might be a good idea to hide this for non enterprise clients? 

You raise a valid point though that the two conditions, feature disabled, and user selecting not to export simulation params may need to be treated differently. Maybe, if the feature is enabled, the checkbox is selected by default, and a warning could be issued if the user deselects the checkbox. Let me check what the business expectation around this would be.

brucenguyen1: "Since the simulator is only meant for enterprise clients": not sure what you meant "simulator" feature. For this particular code, "Process-Simulation-Info-Logic" is located in ApromoreCore (it extracts data from log), the editor is Core, the simulation panel in the editor is Core, so it is public.

lahiru-maha: Oh. I was under the impression the simulation panel in the editor was an EE only feature, hence my comment on the extraction of simulation parameters (via the Process-Simulation-Info-Logic) being a hidden feature, that if enabled would only add the simulation params, but would be useless to non EE users.

brucenguyen1: In case the service is not available for any reason (just as the code can detect that), a little warning/footer note like "*BPMN model will be exported without statistics extracted from logs" makes sense to me, but just go ahead, it's not a big deal.

lahiru-maha: Done. made the changes to add a warning footnote if the feature is disabled

brucenguyen1 (approved):

PR #1622, opened 2022-01-20
Standardising the checkstyle configuration to the root project, so all modules can share a single template

lahiru-maha:
Moving the checkstyle template from the submodules into the root project, so
1.  There is one standard template
2. it can be shared by all submodules
3. template can be imported into development IDEs


franklma (approved):
works as expected

raboczi (approved):
Looks good, makes sense, and minimal.

raboczi (changes requested):
I tried building ApromoreEE/development against ApromoreCore/dev/standardising-checkstyle and the :Apromore-Custom-Plugins:Process-Simulation-Info-Logic:checkstyleMain task failed.  Can you confirm whether this also happens for you?

raboczi (approved):
Looks like I had an older commit; now everything builds correctly.  Approved!

PR #1637, opened 2022-01-24
AP-5643 Toggle attachment display

dreamind:
- Fix replaceAll in safari
- Limit image upload


Janite (approved):
The toggle and image upload limit work as expected.

PR #1639, opened 2022-01-25
AP-5697 Restrict login based on login permission instead of role

Janite:
Changes in this PR:

- Added a portal login permission to the database.
- Updated list of granted authorities to include both roles and permissions, instead of only roles.
- Prevent users from logging in based on whether they have a role with login permission instead of the name of their role.

### DIFF
--- a/Apromore-Core-Components/Apromore-Manager/src/main/java/org/apromore/security/impl/UsernamePasswordAuthenticationProvider.java
+++ b/Apromore-Core-Components/Apromore-Manager/src/main/java/org/apromore/security/impl/UsernamePasswordAuthenticationProvider.java
@@ -223,10 +223,9 @@ public class UsernamePasswordAuthenticationProvider implements AuthenticationPro
     }
 
     /**
-     * Retrieves the correct ROLE or PERMISSION type depending on the access level, where access level is an Integer.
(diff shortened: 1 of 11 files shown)


Apromore-Core-Components/Apromore-Manager/src/main/java/org/apromore/security/impl/UsernamePasswordAuthenticationProvider.java
```diff
@@ -222,7 +223,7 @@ private Authentication authenticatedToken(User user, Authentication original) {
     }
 
     /**
-     * Retrieves the correct ROLE type depending on the access level, where access level is an Integer.
+     * Retrieves the correct ROLE or PERMISSION type depending on the access level, where access level is an Integer.
      * Basically, this interprets the access value whether it's for a regular user or admin.
      *
      * @param access an integer value representing the access of the user
```
raboczi: This doc comment seems to be out of date, since the `access` parameter is a set of `Role`s rather than an integer.

raboczi (commented):
Breaking out PORTAL_LOGIN as a permission definitely tidies things up. There's a doc comment I propose to tidy up before merging it.

raboczi (approved):

PR #1641, opened 2022-01-25
AP-5360 Do not change tree on drag and drop from list

shapon2004:


Janite (approved):
Looks good! Dragging and dropping an item into a folder does not change the which folder is opened, as expected.

PR #1642, opened 2022-01-25
AP-5649-5650 Fix Safari issues for attachments feature

dreamind:


Janite (approved):
Attachments tab layout is fixed in both safari and firefox.

PR #1646, opened 2022-01-27
AP-5678 Change the name of ROLE_ADMIN to ROLE_SUPER_ADMIN

Janite:
This PR has a pair in ApromoreEE: https://github.com/apromore/ApromoreEE/pull/992

Updated the name of ROLE_ADMIN to ROLE_SUPER_ADMIN. Since only the name has been changed, all users which previously had "admin" role, should now have the "super admin" role with the same permissions.

Note: The role map in UserAdminController still contains ROLE_ADMIN as a new admin role will be added after this with less permissions than the current one.


franklma (approved):
LGTM

PR #1647, opened 2022-01-27
AP-5719 Add check for calendar permission in context menu

Janite:
Added a check for calendar permission to prevent users without the permission from seeing the calendar in their context menu.


shapon2004 (approved):
Looks fine to me!

PR #1652, opened 2022-01-27
AP-5313 Right-click menu when multiple files are selected.

shapon2004:


raboczi (changes requested):
The substance of this seems fine.  Could you redo the indentation without tabs?

raboczi (approved):

PR #1654, opened 2022-01-27
AP-5697 Prevent login based on login permission instead of role name

Janite:
This PR is to help prevent login issues due to changes of role names in the database. It includes changes to prevent login based on a login permission instead of a list of role names.

Changes in this PR:
- Added a portal login permission to the database.
- Updated list of granted authorities to include both roles and permissions, instead of only roles.
- Prevent users from logging in based on whether they have a role with login permission instead of the name of their role.


raboczi (approved):
Same changes as went into the development branch.

PR #1657, opened 2022-01-28
AP-5724 Fix tab refresh issue in the BPMN editor props panel

dreamind:


Janite (approved):
The simulation parameters appear when switching from an end node to a task node as expected.

PR #1669, opened 2022-01-31
AP-5724 Fix tab refresh issue in the BPMN editor props panel

dreamind:


Janite (approved):
The simulation tasks tab is appearing as expected when switching from the end event to a task node.

PR #1670, opened 2022-01-31
AP-5313 Right-click menu when multiple files are selected (New Changes)

shapon2004:


Apromore-Core-Components/Apromore-Portal/src/main/resources/zk-label_ja.properties
```diff
@@ -328,6 +328,7 @@ plugin_file_manageDataPipelines_text = データパイプラインの管理
 plugin_analyze_title_text = 分析する
 plugin_redesign_title_text = 再設計する
 plugin_analyze_animateLogs_text = ログをアニメートする
+plugin_analyze_animateLog_text = Animate log
```
Janite: You can use the same translation for Animate logs for this ログをアニメートする. The translation in this case is the same with or without plurals.

Janite (approved):
The menu items show as expected when multiple files are selected.

PR #1672, opened 2022-01-31
AP-5754 Fix incorrectly triggered calendar change event

dreamind:


Janite (approved):
Adding a calendar no longer triggers PD to refresh.

PR #1673, opened 2022-01-31
AP-5754 Fix incorrectly triggered calendar change event

dreamind:


Janite (approved):
PD refresh is no longer triggered when a calendar is created. Refresh still occurs when a calendar is applied.

PR #1675, opened 2022-02-01
AP-5679 Implement new admin role

Janite:
1.  Added a new admin role which has the following permissions:
- Login permission
- View users
- Edit users
- Edit groups
- Edit roles

2. Updated access to user and group management to be permission based instead of role based.
3. Remove unused isCurrentUserAdmin() method in MainController


franklma (approved):
LGTM

PR #1677, opened 2022-02-01
AP-5754 Fine tune calendar-log sync for delete/change events

dreamind:


Janite (approved):
Deleting an unrelated custom calendar no longer prompts PD refresh. Deleting the calendar applied to the log opened in PD prompts refresh.

PR #1680, opened 2022-02-01
AP-5680 Implement data engineer role

Janite:
This PR adds the "Data Engineer" role which has the following permissions:
- Login permission
- Create data pipelines
- Manage data pipelines


franklma (approved):
LGTM

PR #1691, opened 2022-02-03
Adding role as a perspective when importing a csv log - pt1

lahiru-maha:
- Adding role as a perspective when importing a csv log, in order to enable roles in PD
- Some code refactoring to
  - move common methods to an AbstractLogImporter class
  - remove unnecessary LogEventModelExt dto
- adding unit tests


Apromore-Custom-Plugins/Log-Importer-Logic/src/main/java/org/apromore/service/logimporter/services/legacy/AbstractLogImporter.java
```diff
@@ -0,0 +1,183 @@
...
+        tracesHistory.forEach(
+            /*
+             * Java 8 (k, v) -> { v.sort(new XEventComparator()); xLog.add(v); }
+             */
+            new java.util.function.BiConsumer<String, XTrace>() {
+                public void accept(String k, XTrace v) {
+                    v.sort(new XEventComparator());
+                    xLog.add(v);
+                }
+            });
```
raboczi: Why isn't the Java 8 syntax being used?

franklma (approved):
Need to add a mapping from extension name "org:role" to "Role" inside the APMLog module as per discussion with @chiichang. Other than that, LGTM.

chiichang (approved):

Apromore-Custom-Plugins/Log-Importer-Logic/src/main/java/org/apromore/service/logimporter/services/legacy/AbstractLogImporter.java
lahiru-maha: This was from a copy/paste while doing a bit of refactoring. I've fixed this up now, as part of addressing sonar cloud code smells. Will address the remaining Java 8 todo comments across the code in a subsequent PR

lahiru-maha:
Will address the remaining code smell with the next round of PRs (which would involve code refactoring) that I raise for this

PR #1692, opened 2022-02-03
Setting the display time unit based on the value of the time duration.

lahiru-maha:
Setting the display time unit based on the value of the time duration.

If duration <= 60 seconds, time unit is seconds
If duration > 60 seconds and <= 3600, time unit is minutes
If duration > 3600 time unit is hours

The display duration will ALWAYS be converted to seconds for now, as both the BPMN editor and the simulator always operate in seconds.

### DIFF
--- a/Apromore-Custom-Plugins/Process-Simulation-Info-Logic/src/main/java/org/apromore/processsimulation/service/SimulationInfoService.java
+++ b/Apromore-Custom-Plugins/Process-Simulation-Info-Logic/src/main/java/org/apromore/processsimulation/service/SimulationInfoService.java
@@ -146,23 +146,22 @@ public class SimulationInfoService {
         final SimulationData simulationData) {
 
         List<Element> taskList = null;
-        if (simulationData.getDiagramNodeIDs() != null) {
-            taskList = simulationData.getDiagramNodeIDs().stream()
-                .map(nodeId -> {
-                    double durationMillis = simulationData.getDiagramNodeDuration(nodeId);
-                    TimeUnit timeUnit = getDisplayTimeUnit(durationMillis);
-                    return Element.builder()
-                        .elementId(nodeId)
-                        .distributionDuration(Distribution.builder()
-                            .type(DistributionType.valueOf(
-                                config.getDefaultDistributionType().toUpperCase(DOCUMENT_LOCALE)))
-                            .arg1(getDisplayTimeDuration(durationMillis).toString())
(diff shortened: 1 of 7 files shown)


Apromore-Custom-Plugins/Process-Simulation-Info-Logic/src/main/java/org/apromore/processsimulation/service/SimulationInfoService.java
```diff
@@ -224,6 +230,34 @@ private void deriveGatewayProbabilities(
...
+
+        if (timeDurationMillis <= 60000) {
+            timeUnit = TimeUnit.SECONDS;
+        } else if (timeDurationMillis > 60000 && timeDurationMillis <= 3600000) {
+            timeUnit = TimeUnit.MINUTES;
+        } else if (timeDurationMillis > 3600000) {
+            timeUnit = TimeUnit.HOURS;
+        } else {
+            timeUnit = TimeUnit.valueOf(config.getDefaultTimeUnit().toUpperCase(DOCUMENT_LOCALE));
+        }
```
raboczi, line 244: I feel like this could be simplified to just:
```
timeUnit = timeDurationMillis <= 60000   ? TimeUnit.SECONDS :
           timeDurationMillis <= 3600000 ? TimeUnit.MINUTES :
                                           TimeUnit.HOURS;
```

raboczi (commented):
There's nothing in here I'd object to merging.  One comment about the actual time unit selection logic I'd like you to consider before I go ahead and approve.

Apromore-Custom-Plugins/Process-Simulation-Info-Logic/src/main/java/org/apromore/processsimulation/service/SimulationInfoService.java
lahiru-maha: I'm not a huge fan of using the ternary operator, as it reduces readability. I typically use it sparingly, where there's only 2 branching possibilities

raboczi: Just as long as you've considered collapsing that logic and whether that final unreachable branch deserves to continue existing, that'll all I was really interested in.

raboczi (approved):

PR #1697, opened 2022-02-04
AP-5689 Improvement in Submenu

shapon2004:
Removed also unused sub-menu from menu.json


raboczi (changes requested):
Please break PopupLogSubMenuController into four subclasses (e.g. DiscoverPopupLogSubMenuController, DashboardPopupLogSubMenuController, LogFilterPopupLogSubMenuController, and CalendarPopupLogSubMenuController) to eliminate all the subMenuId conditional logic.

shapon2004:
It has been changed accordingly!

Apromore-Core-Components/Apromore-Portal/src/main/java/org/apromore/portal/controller/PopupLogSubMenuController.java
```diff
@@ -0,0 +1,221 @@
...
+                    calendarPlugin.setSimpleParams(attrMap);
+                    calendarPlugin.execute(getPortalContext());
+                } catch (Exception e) {
+                    LOGGER.error(Labels.getLabel("portal_failedLaunchCustomCalendar_message"), e);
+                    Messagebox.show(Labels.getLabel("portal_failedLaunchCustomCalendar_message"));
+                }
+            });
+            menuPopup.appendChild(item);
+        }
+    }
```
raboczi: Move to CalendarPopupLogSubMenuController.

Apromore-Core-Components/Apromore-Portal/src/main/java/org/apromore/portal/controller/PopupLogSubMenuController.java
```diff
@@ -0,0 +1,221 @@
...
+    private void viewExistingDashboard(UserMetadataSummaryType um) {
+        try {
+            Sessions.getCurrent().setAttribute("logSummaries", Collections.singletonList(logSummaryType));
+            Sessions.getCurrent()
+                .setAttribute("userMetadata_dash", userMetaDataUtilService.getUserMetaDataById(um.getId()));
+            Clients.evalJavaScript("window.open('dashboard/index.zul')");
+        } catch (Exception e) {
+            LOGGER.error("Error in showing the Dashboard", e);
+        }
+    }
```
raboczi: Move to DashboardPopupLogSubMenuController.

Apromore-Core-Components/Apromore-Portal/src/main/java/org/apromore/portal/controller/PopupLogSubMenuController.java
```diff
@@ -0,0 +1,221 @@
...
+        item.setAttribute(USER_META_DATA, um);
+        item.addEventListener(ON_CLICK, event -> {
+            try {
+                if (SUB_MENU_FOR_PD.equals(event.getTarget().getAttribute(SUB_MENU_FOR))) {
+                    viewPdOrLogFilter((UserMetadataSummaryType) event.getTarget().getAttribute(USER_META_DATA), false);
+                } else if (SUB_MENU_FOR_LOG_FILTER.equals(event.getTarget().getAttribute(SUB_MENU_FOR))) {
+                    viewPdOrLogFilter((UserMetadataSummaryType) event.getTarget().getAttribute(USER_META_DATA), true);
+                } else if (SUB_MENU_FOR_DASHBOARD.equals(event.getTarget().getAttribute(SUB_MENU_FOR))) {
+                    viewExistingDashboard((UserMetadataSummaryType) event.getTarget().getAttribute(USER_META_DATA));
+                }
```
raboczi: Shouldn't need this conditional code, because the subclass of PopupLogSubMenuController will know whether it's PD, log filter, dashboard, etc.

Apromore-Core-Components/Apromore-Portal/src/main/java/org/apromore/portal/controller/PopupLogSubMenuController.java
```diff
@@ -0,0 +1,221 @@
...
+            } catch (Exception ex) {
+                LOGGER.error("Error in forwarding the request", ex);
+            }
+        });
+        item.setVisible(visibleOnLoad);
+        popup.appendChild(item);
+    }
+
+    private void viewPdOrLogFilter(UserMetadataSummaryType um, boolean edit) {
```
raboczi: Make this protected so that it can be called from its subclasses instead.

Apromore-Core-Components/Apromore-Portal/src/main/java/org/apromore/portal/controller/PopupLogSubMenuController.java
shapon2004: It is not related with sub-class. It is related with event fired from the front-end and then captured from the back-end to distinguish type of event has been fired.

Apromore-Core-Components/Apromore-Portal/src/main/java/org/apromore/portal/dialogController/PopupMenuController.java
```diff
@@ -399,81 +407,75 @@ private void viewExistingDashboard() {
...
-            MenuConfig menuConfig = menuConfigLoader.getMenuConfig(subMenuId);
-            if (menuConfig.getGroups().isEmpty()) {
+            if (pluginAvailable(PluginCatalog.PLUGIN_DISCOVER_MODEL)) {
+                new DiscoverPopupLogSubMenuController(this, mainController, popup,
+                    (LogSummaryType) selections.iterator().next());
+            }
+            } catch (Exception e) {
+            LOGGER.error(FAILED_LOAD_SUB_MENU, e);
+        }
+    }
```
raboczi: There's a chunk of repeated code in here common with `addLogFilterSubMenuItem`, `addDashboardSubMenuItem` etc.  There's probably a clever way to pass the constructor as an argument, specify the construct as a generic, or move it to one of the subclasses, but I think it's tricky enough that I'm willing to let it slide unless you're ambitious to figure it out.

raboczi (changes requested):
Now that the subclasses of PopupLogSubMenuController exist, there's a little bit more code that can be moved out into them.

Apromore-Core-Components/Apromore-Portal/src/main/java/org/apromore/portal/controller/PopupLogSubMenuController.java
raboczi: Should the subclasses ever receive and have to handle an event for which they are not the SUB_MENU_FOR?  I assumed not, but I may have missed how this works.

raboczi (approved):
I reckon it's good enough to merge at this point.  Thanks for indulging my fussiness!

PR #1701, opened 2022-02-07
AP-5360 the tree should not change. It should neither expand nor rollup.

shapon2004:


Apromore-Core-Components/Apromore-Portal/src/main/java/org/apromore/portal/dialogController/BaseListboxController.java
```diff
@@ -793,8 +793,7 @@ public void onEvent(Event evt) throws Exception {
             switch (((Integer) evt.getData())) {
               case Messagebox.YES:
                 deleteElements(mainController);
-                mainController.loadWorkspace();
-                refreshContent();
+                mainController.reloadSummariesWithOpenTreeItems(mainController.getNavigationController().getAllOpenFolderItems());
```
Janite: There is a private refreshWorkspace() method in BaseListboxController which contains the same code as this line. Perhaps we can reuse that method in this class to avoid duplication.

shapon2004: private refreshWorkspace() actually contains only same line
"mainController.reloadSummariesWithOpenTreeItems(mainController.getNavigationController().getAllOpenFolderItems());"

shapon2004: OK... Using That!

Janite (approved):
The folder does not change when pasting into to a parent or child folder or when renaming or deleting a folder.

PR #1704, opened 2022-02-08
AP-5793 Unable to save file name when user uploads a log with more than 100 chars

shapon2004:
Applied same pattern in both places (Property file and Java class):
^[a-zA-Z0-9, &\u0080-\u9fff\u005B\u005D\._\+\-\(\)%]{1,100}$


franklma (approved):
LGTM

PR #1709, opened 2022-02-09
Exporting sim params resource counts grouped by role

lahiru-maha:
- Exporting resources by role
- Fixing bug AP-5854, where export was not possible if the view in PD was set to BPMN


Apromore-Custom-Plugins/Process-Discoverer-Portal-Plugin/src/main/java/org/apromore/plugin/portal/processdiscoverer/PDAnalyst.java
```diff
@@ -669,34 +671,133 @@ public void updateLog(PLog pLog, APMLog apmLog) throws Exception {
...
+            .stream()
+            .filter(Activity.class::isInstance)
+            .collect(Collectors.toMap(
+                //key --> node Id
+                bpmnNode -> bpmnNode.getId().toString(),
+                //value --> role
+                bpmnNode -> {
+                    String nodeLabel = bpmnNode.getLabel();
+                    return ObjectUtils.isEmpty(nodeLabel)
+                        ? SimulationData.DEFAULT_ROLE : activityNameToRoleMap.get(nodeLabel);
```
brucenguyen1: For safety reason, should be: return ObjectUtils.isEmpty(nodeLabel) || !activityNameToRoleMap.containsKey(nodeLabel) ? SimulationData.DEFAULT_ROLE : activityNameToRoleMap.get(nodeLabel).

brucenguyen1 (approved):

Apromore-Custom-Plugins/Process-Discoverer-Portal-Plugin/src/main/java/org/apromore/plugin/portal/processdiscoverer/PDAnalyst.java
```diff
@@ -669,34 +671,133 @@ public void updateLog(PLog pLog, APMLog apmLog) throws Exception {
...
+                /*
+                 * Merge function --> Ignore duplicate keys
+                 * role1 --> The role currently in the map against the key (activity name)
+                 * role2 --> The new role that matches an existing key (activity name)
+                 *
+                 * An activity can be performed by multiple resources, each with the same or different roles
+                 * We will treat the first role of an activity as the primary role for that activity, and ignore
+                 * the rest.
+                 */
+                (role1, role2) -> role1
```
brucenguyen1: If there is a first activity instance with activity "A" but no role, then "A" will be mapped to DEFAULT_ROLE, no matter the succeeding instances where "A" is associated with a non-empty role. This is because DEFAULT_ROLE is always chosen in the merge function above.

More accurately, for an activity "A", we should return the first non-empty role associated with "A" in the sequence of activity instances of "A".

brucenguyen1: The merge function could be: (role1, role2) -> !role1.equals(DEFAULT_ROLE) ? role1 : role2.

Apromore-Custom-Plugins/Process-Discoverer-Portal-Plugin/src/main/java/org/apromore/plugin/portal/processdiscoverer/PDAnalyst.java
```diff
@@ -669,34 +671,133 @@ public void updateLog(PLog pLog, APMLog apmLog) throws Exception {
...
+            .forEach(activityInstance -> {
+                String role = activityInstance.getAttributeValue(Constants.ATT_KEY_ROLE);
+
+                if (role == null || role.isEmpty()) {
+                    role = SimulationData.DEFAULT_ROLE;
+                }
+                Set<String> resources = rolesToResourcesMap.get(role);
+                if (resources == null) {
+                    resources = new HashSet<>();
+                }
```
brucenguyen1, line 723: Can be shorter: rolesToResourcesMap.getOrDefault(role, new HashSe<>());

Apromore-Custom-Plugins/Process-Discoverer-Portal-Plugin/src/main/java/org/apromore/plugin/portal/processdiscoverer/PDAnalyst.java
```diff
@@ -669,34 +671,133 @@ public void updateLog(PLog pLog, APMLog apmLog) throws Exception {
...
+                resources.add(activityInstance.getResource());
+                rolesToResourcesMap.put(role, resources);
+            });
+
+        return rolesToResourcesMap.entrySet().stream()
+            .collect(Collectors.toMap(Map.Entry::getKey, stringSetEntry -> stringSetEntry.getValue().size()));
```
brucenguyen1: What happens when activity instances could have org:role attribute (e.g. Role1) but no org:resource, Role1 should be mapped to DEFAULT_RESOURCE but it is omitted.
Maybe we should make it consistent rule: for any activity instance, no org:role is considered as DEFAULT_ROLE, no org:resource is considered as DEFAULT_RESOURCE and it could be helpful to avoid if role is not DEFAULT_ROLE then...else in other code that uses this mapping?

If the rule is consistent above, we can make the code easy:
- No need to filter out activity instances without org:resource attribute.
- resource = activityInstance.getAttributes().containsKey(Constants.ATT_KEY_RESOURCE) ? activityInstance.getResource() : DEFAULT_RESOURCE
- role = activityInstance.getAttributes().containsKey(Constants.ATT_KEY_ROLE) ? activityInstance.getAttributeValue(Constants.ATT_KEY_ROLE) : DEFAULT_ROLE
- Add resource to the set of role's resources.
- No need if (role is DEFAULT_ROLE) then....else.... in other code.

PR #1711, opened 2022-02-09
AP-5858 Multiple calendars associated with a log but only displays one.

shapon2004:


dreamind:
@shefali316 @shapon2004 I'm not sure if we should allow multiple calendars to be associated with one log. I think the current assumption is that one log can only be associated with one calendar.

shapon2004:
@dreamind , We are just showing all the calendars just like we have in calendar popup window along with Calendar associated with the log. There is no modification on the logic we already have but just redirecting to particular calendars.

shapon2004:
@dreamind , I have also mentioned same thing into https://apromore.atlassian.net/browse/AP-5479

dreamind (approved):

PR #1720, opened 2022-02-11
bugfix/AP-5865-incorrect-number-of-models-selected

Janite:
Fixed an issue where multiple VersionSummaryType objects would be added to the selection when selecting a process if
1. no process version was selected and
2. there were multiple VersionSummaryType objects with the same version number as the latest version for the process.

Added a check to only add the first VersionSummaryType in the "MAIN" branch.


franklma (approved):
LGTM

PR #1727, opened 2022-02-14
AP-5880 Fix calendar unlink issue

dreamind:


Janite (approved):
Tested locally, PD refreshes and the values are restored when restoring a calendar.

PR #1729, opened 2022-02-15
AP-5889 Associate new calendars with a user

Janite:
- Added an "owner" column to custom_calendar which contains the id of the user who owns the calendar.
- When a calendar is created, the "owner" column will be set to the id of the user who created the calendar.
- When a calendar is created, the "created_by" column will be set to the username of the user who created the calendar.


franklma (approved):
LGTM

PR #1731, opened 2022-02-15
AP-5891 Transfer calendar ownership when assets are transferred

Janite:
Update the calendar owner when assests are transferred.


Janite (commented):
Sonarcloud is currently down for scheduled maintenance. Will re-run the build check once it's up again.

franklma (approved):
LGTM

PR #1732, opened 2022-02-16
AP-5892 Users can see only owned and applied calendars

Janite:
- Added support to get a list of owned calendars
- Updated the calendar list to show only owned calendars and the calendar applied to the selected log. This should apply to both the list shown in calendar management and the calendar list for the popup submenu.
- Updated default calendar name to include the user who created it.


shapon2004 (approved):
Looks Fine!

PR #1735, opened 2022-02-16
AP-5890 Remove calendars not associated with a user

Janite:
Remove pre-existing calendars with no created_by or owner field.


franklma (approved):
LGTM

PR #1736, opened 2022-02-16
AP-5887 Scroll bar missing in the Activity Inspector

shapon2004:


Janite (approved):
This is the same as what we have in the case inspector and case variant inspector. Scroll bar shows as expected.

PR #1741, opened 2022-02-18
AP-5929 Fix publish model issues

Janite:
- Skip autosave if the model is opened in view mode using a publish model link. This fixes an issue where opening a publish model view link throws an error due to not having an edit session id parameter.
- Add a check requiring users to save a model before publishing it.


franklma (approved):
LGTM

PR #1754, opened 2022-02-22
Bugfix/ap 5858 click on a calendar from the list will be applied immediately

shapon2004:
It should not edit the calendar.


Janite (approved):
Works as expected
- Calendars are applied automatically for users with editor and owner access
- Users with view access recieve an error message when attempting to create a new calendar or apply a calendar
- Selecting a calendar with a tick mark does nothing

PR #1766, opened 2022-02-23
AP-4801 Clear search on empty query

dreamind:


Janite (approved):
Deleting all characters from the searchbox and hitting enter returns all results as expected.

PR #1767, opened 2022-02-24
AP-5976 Restrict publish model access based on publish model permission

Janite:
- Added a publish model permission. Only the Super Admin, Analyst and Designer roles have publish model permission by default.
- Show/hide publish model based on publish model permission.


franklma (approved):
LGTM

PR #1773, opened 2022-02-26
AP-5922 Mulitple session handle from Portal

shapon2004:


shapon2004:
To resolve the session related issue, following options has been implemented:

1) Multiple portalContext will be stored into session for multiple Portal tabs.(ex: Sessions.getCurrent().setAttribute("portalContext_" + Desktop_Id, portalContext))

2) To access the portalContext, other tabs like dashboard or PD will have to include reference of the portal into Url (ex: dashboard/index.zul?REFER_ID=Desktop_Id)

3) During creation of other tabs(excluding Portal), a new ZK desktop attribute has been added to set related parent portal (ex:  desktop.setAttribute("PORTAL_REF_ID",Desktop_Id))

4) To retrieve related portalContext: We call first desktop.getAttribute("PORTAL_REF_ID") to retrieve parent portal reference and thenusing the reference we identify the related portalContext from session (ex: Sessions.getCurrent().getAttribute("portalContext_"+Desktop_Id))

5) For portal, we have added a listner into Maincontroller to cleanup the sesion portalContext related object.

6) All the class/object will get the portalContext from PortalContextHolder.getActivePortalContext() static method. This method has the mechanism to retrieve the portalContext from current tabs.

Apromore-Core-Components/Apromore-Portal/src/main/java/org/apromore/portal/context/PluginPortalContext.java
```diff
@@ -132,11 +131,11 @@ public FolderType getCurrentFolder() {
...
     @Override
     public UserType getCurrentUser() {
+        //If parent window is closed, then below code block gives nullpointer
         //return UserSessionManager.getCurrentUser();
-    	Desktop desktop = mainController.getDesktop();
-        Session session = desktop.getSession();
-        UserType userType = (UserType) session.getAttribute(UserSessionManager.USER);
-    	return userType;
+        //Desktop desktop = mainController.getDesktop();
+        //Session session = desktop.getSession();
```
raboczi: Reasonable to remove this commented-out code?

Apromore-Plugins/plugin-core/portal/api/src/main/java/org/apromore/plugin/portal/PortalUrlWrapper.java
```diff
@@ -0,0 +1,42 @@
...
+
+    }
+    public static String getUrlWithReference(String url) {
+        if (url == null || url.isEmpty()) {
+            return "";
+        } else if (url.contains("?") || url.contains("=")) {
+            url += "&REFER_ID=" + Executions.getCurrent().getDesktop().getId();
+        }else{
+            url += "?REFER_ID=" + Executions.getCurrent().getDesktop().getId();
+        }
```
raboczi: Perhaps java.net.URI should be used here to incorporate some validation?  The conditional check would become something like `URI uri = new URI(url); if (uri.getQuery() == null) { ...`.  It'd be a bit verbose, but the output could be composed using a URI instance as well.

raboczi (changes requested):
There seem to be tabs introduced into the source code again, e.g. in the new PortalContextHolder.

My preliminary testing turned up an exception when opening Dashboard tabs.  I should be able to demonstrate that to you if needed.

Apromore-Core-Components/Apromore-Portal/src/main/java/org/apromore/portal/context/PluginPortalContext.java
shapon2004: Yeah. I am removing it.

raboczi: Passes for me now.

raboczi (approved):
This seems to address the defect and is fairly tidy.  Approved!

PR #1775, opened 2022-02-28
Feature/ap 5879

dreamind:
Prove a new button to show/hide all comments and should be positioned after the one for showing/hiding attachments.


Janite (approved):
Comment toggle works as expected.

PR #1776, opened 2022-03-01
AP-5980-5983-5985 Cost overlay improvements

dreamind:


Janite (changes requested):
Looks good. There is one point in the card about repositioning the window to be in the center of the screen, like for calendars. But the window is not in the center.

Requested a change to move the Currency label to a properties file.

Apromore-Custom-Plugins/Process-Discoverer-Portal-Plugin/src/main/resources/static/processdiscoverer/zul/costTable.zul
```diff
@@ -41,11 +42,13 @@
       </template>
     </listbox>
     <div height="40px" hflex="1" style="padding:4px; padding-left: 10px;">
+      <label style="margin-right: 6px">Currency</label>
```
Janite: We can move the Currency label to a property file. The translation for it should be the same as Currency in BPMN Editor which is 通貨

Janite (approved):
LGTM!

PR #1778, opened 2022-03-01
AP-6043 Create a permission for Manage access rights

Janite:
- Create a permission for Manage access rights to decouple its access from "User management".
- Give the admin and super admin roles Manage access rights permission by default.


franklma (approved):
LGTM

PR #1784, opened 2022-03-02
feat: added the entry points for the PPM plugin

mayankgituni:
Added the predictive monitoring plugin project to ApromoreEE and added the entry point from the Portal.


Janite (approved):
LGTM! The predictor trainer plugin appears in the analyze menu and log right-click menu as expected.

PR #1787, opened 2022-03-03
hotfix/v8.2 Add a check for empty permission names when adding authorities

Janite:
An IllegalArgumentException is thrown when attempting to create a Granted Authority with an empty string as the name. Added a check for non-empty strings before adding the authority.


franklma (approved):
LGTM

PR #1788, opened 2022-03-03
Feature/ap 6010/xes to parquet converter

mayankgituni:
Added the XES to parquet converter service in the Apromore-Core-Plugin's Parquet-Log-Exporter plugin. This service takes the ID of the event log and converts the APMLog into parquet and then streams the file to a filesystem i.e S3 or LocalFS.


shapon2004 (approved):
Looks Good!

PR #1789, opened 2022-03-03
AP-5533 Refine custom properties

dreamind:


Janite (approved):
Looks good!
- [x] BPMN Editor opens in edit and view mode
- [x] Model can be viewed with a publish model link
- [x] Model name appears in the properties panel
- [x] Metadata properties can be added to data objects, data stores, pools and lanes

PR #1800, opened 2022-03-09
Feature/ap 6008/view prediction UI

mayankgituni:
This PR adds the ability to fetch the list of selected event logs from the portal context in an easier way. We have introduced a new method i.e. getSelectedArtifacts to get the selected BPMN models and event logs.

### DIFF
--- a/Apromore-Core-Components/Apromore-Portal/src/main/java/org/apromore/portal/context/PluginPortalSelection.java
+++ b/Apromore-Core-Components/Apromore-Portal/src/main/java/org/apromore/portal/context/PluginPortalSelection.java
@@ -1,3 +1,21 @@
+/**
+ * #%L
+ * This file is part of "Apromore Enterprise Edition".
+ * %%
+ * Copyright (C) 2019 - 2021 Apromore Pty Ltd. All Rights Reserved.
+ * %%
+ * NOTICE:  All information contained herein is, and remains the
+ * property of Apromore Pty Ltd and its suppliers, if any.
+ * The intellectual and technical concepts contained herein are
+ * proprietary to Apromore Pty Ltd and its suppliers and may
+ * be covered by U.S. and Foreign Patents, patents in process,
+ * and are protected by trade secret or copyright law.
+ * Dissemination of this information or reproduction of this
+ * material is strictly forbidden unless prior written permission
+ * is obtained from Apromore Pty Ltd.
+ * #L%
+ */
+
 package org.apromore.portal.context;
 
 import java.util.ArrayList;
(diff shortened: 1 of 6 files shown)


Apromore-Core-Components/Apromore-Portal/src/main/java/org/apromore/portal/context/PluginPortalSelection.java
```diff
@@ -0,0 +1,68 @@
...
+ * property of Apromore Pty Ltd and its suppliers, if any.
+ * The intellectual and technical concepts contained herein are
+ * proprietary to Apromore Pty Ltd and its suppliers and may
+ * be covered by U.S. and Foreign Patents, patents in process,
+ * and are protected by trade secret or copyright law.
+ * Dissemination of this information or reproduction of this
+ * material is strictly forbidden unless prior written permission
+ * is obtained from Apromore Pty Ltd.
+ * #L%
+ */
```
Janite: This license header should be for ApromoreCore.

Apromore-Custom-Plugins/Parquet-Exporter-Plugin/src/main/java/org/apromore/plugin/parquet/export/service/XesToParquetConverterService.java
```diff
@@ -1,3 +1,21 @@
...
+ * The intellectual and technical concepts contained herein are
+ * proprietary to Apromore Pty Ltd and its suppliers and may
+ * be covered by U.S. and Foreign Patents, patents in process,
+ * and are protected by trade secret or copyright law.
+ * Dissemination of this information or reproduction of this
+ * material is strictly forbidden unless prior written permission
+ * is obtained from Apromore Pty Ltd.
+ * #L%
+ */
+
```
Janite: This license header should be for ApromoreCore.

Janite (changes requested):
The license header for ApromoreEE has been added in a few places instead of the ApromoreCore license header.

Apromore-Plugins/plugin-core/portal/api/src/main/java/org/apromore/plugin/portal/PortalContexts.java
```diff
@@ -73,4 +77,11 @@ public static void removePortalContextReference(Desktop desktop) {
             LOGGER.error("Error in removing portalContext", ex);
         }
     }
+
+    public static PageDefinition getPageDefination(String path) throws IOException {
```
Janite: Can we rename this method to getPageDefinition?

Janite (changes requested):
There are a couple of code smells raised by sonarcloud about unused imports and returning null.

Janite (approved):

PR #1801, opened 2022-03-09
Bug fixes for 3 defects in the BPMN Editor & Sim param export

lahiru-maha:
- AP-6609: Changing the labelling from Resources to "Resource pools" in the BPMN Editor to help minimise the confusion between Resources and Roles
- AP-6071: Reverting the start time back to epoch time, which is what is required by the BPMN Editor (as opposed to local date time) and setting the ReferId from the desktop so the portalContext is available on that view
- AP-5813: Renaming "Default Resource" to "Default" as "Resource" is a keyword in the dashboard that prevents the loading of the resources graph on a simulated log's dashboard


raboczi (approved):
My only comment is that the GitHub differencer didn't cope with bpmneditor.js and bpmn-modeler.development.js (best guess, your editor changed the line endings) so I can't actually see your changes to those files.

shapon2004 (approved):

chiichang (approved):

PR #1811, opened 2022-03-14
AP-6135 Portal: owner name is not flushed right anymore

shapon2004:


raboczi (approved):
I kind of wish we weren't using that "view" cookie, but it's preexisting mechanism so it'd be out of scope to fix it as part of this feature.

PR #1814, opened 2022-03-15
AP-6163: Fix SonarCloud XXE warning

raboczi:
The parser which reads the XSL transformation used to filter out script injections inside uploaded BPMN documents wasn’t being configured to prevent XXE attacks.  This isn’t a practical issue since that document is loaded from the classpath rather than from anywhere a user could tinker with it, but it’s easy enough to fix and keep SonarCloud happy.

This is pure refactoring, and causes no change in behavior.


franklma (approved):
LGTM

PR #1817, opened 2022-03-15
AP-5822 Update label and icon for View/Edit model plugin

Janite:
- Updated the label and icon of Edit model. If a user does not have edit model permission, the label is set to "View model" and the model icon with no pencil is used.
- Updated misaligned japanese labels.


shapon2004 (approved):
Looks fine.

PR #1822, opened 2022-03-16
Use the 24/7 timetable if no calendar set in PD

lahiru-maha:
- Defaulting to the Generic 24/7 calendar model for the arrival timetable if no timetable present from PD. This makes it consistent as PD also defaults to the Generic 24/7 calendar
- If a calendar model is available, select this as the active calendar, but also include the 24/7 timetable for convenience
- Removed the previously hardcoded defaults
- Updated and added tests


raboczi (approved):
I don't have any suggestions that would improve this.

Apromore-Custom-Plugins/Process-Simulation-Info-Logic/src/main/java/org/apromore/processsimulation/service/SimulationInfoService.java
```diff
@@ -134,6 +137,18 @@ public ProcessSimulationInfo transformToSimulationInfo(
...
+        if (simulationData.getCalendarModel() == null) {
+            // If no calendar, default to a 24 x 7 calendar
+            activeCalendar = calendarService.getGenericCalendar();
+        } else {
+            activeCalendar = simulationData.getCalendarModel();
+        }
+
+        return activeCalendar;
+    }
+
```
brucenguyen1: Do we really need this getActiveCalendar with null processing? It could be better to put a constraint NonNull for calendarModel field in SimulationData rather than checking for error here.  It helps detect errors early and fail fast instead of silently passing it like this way. SimulationData represents the contract between simulation data provider (e.g. PDAnalyst) and consumer (e.g. SimulationInfoService). Pls consider adding more constraints such as Non-Null, Non-Empty for other fields if we think it can help enforce this contract.

Apromore-Custom-Plugins/Process-Simulation-Info-Logic/src/main/java/org/apromore/processsimulation/service/SimulationInfoService.java
```diff
@@ -203,26 +213,44 @@ private void deriveTaskInfo(
...
-                            .toUpperCase(DOCUMENT_LOCALE)))
-                    .fromTime(config.getDefaultTimetable().get(CONFIG_DEFAULT_TIMESLOT_FROM_TIME))
-                    .toTime(config.getDefaultTimetable().get(CONFIG_DEFAULT_TIMESLOT_TO_TIME))
-                    .build()))
-                .build()));
+        final ProcessSimulationInfo.ProcessSimulationInfoBuilder builder,
+        final CalendarModel calendarModel) {
+
+        if (calendarModel.getName().equals("Generic 24/7")) {
+            builder.timetables(List.of(createTimetable(
```
brucenguyen1: Hard-coded "Generic 24/7"?

brucenguyen1: We can support this need by using AbsoluteCalendarModel, this is a 24x7 calendar type, adding isAllDayAllTime method to CalendarModel?

Apromore-Custom-Plugins/Process-Simulation-Info-Logic/src/main/java/org/apromore/processsimulation/config/SimulationInfoConfig.java
```diff
@@ -36,6 +36,7 @@
 @ConfigurationProperties(prefix = "process-simulation-info-export")
```
brucenguyen1, line 33: Can we make SimulationConfig immutable class decorated with ConfigurationProperties (i.e. read-only properties)?

Apromore-Custom-Plugins/Process-Simulation-Info-Logic/src/main/java/org/apromore/processsimulation/service/SimulationInfoService.java
```diff
@@ -203,26 +213,44 @@ private void deriveTaskInfo(
...
+                calendarModel, config.getTimetable().get(CONFIG_CUSTOM_ID_KEY), true)));
+        } else {
+            builder.timetables(
+                List.of(
+                    createTimetable(calendarModel, config.getTimetable().get(CONFIG_CUSTOM_ID_KEY), true),
+                    createTimetable(calendarService.getGenericCalendar(),
+                        config.getTimetable().get(CONFIG_DEFAULT_ID_KEY), false)
+                ));
```
brucenguyen1: Can we enforce the rule there is only one default timetable in this list?

lahiru-maha: The requirement was to also provide the 24/7 calendar as well if a custom calendar was specified.

Apromore-Custom-Plugins/Process-Simulation-Info-Logic/src/main/java/org/apromore/processsimulation/service/SimulationInfoService.java
```diff
@@ -277,7 +305,7 @@ private void deriveTimetable(
                         .id(resourceId)
                         .name(resourceName)
                         .totalAmount(roleToResourceCount.getValue())
-                        .timetableId(config.getDefaultTimetable().get(CONFIG_DEFAULT_ID_KEY))
+                        .timetableId(config.getTimetable().get(CONFIG_CUSTOM_ID_KEY))
```
brucenguyen1: config.getTimetable().get(CONFIG_CUSTOM_ID_KEY): this can be null if the timetable map was not initialized with CONFIG_CUSTOM_ID_KEY, depending on the values put in the properties file (see SimulationInfoConfig), which is not safe because the initialization of the timetable map and CONFIG_CUSTOM_ID_KEY are separated. The same for the defaultResource map.

lahiru-maha: The custom ID is part of the timetable properties map and the configuration is initialised at the time it gets autowired.

```
process-simulation-info-export.timetable.id=DEFAULT_TIMETABLE
process-simulation-info-export.timetable.custom-id=CUSTOM_TIMETABLE
process-simulation-info-export.timetable.timeslot-name=Default Timeslot
```

Apromore-Custom-Plugins/Process-Simulation-Info-Logic/src/main/java/org/apromore/processsimulation/config/SimulationInfoConfig.java
lahiru-maha: Don't think spring allows that. It requires setters to initialise the configuration at the time of auto wiring

Apromore-Custom-Plugins/Process-Simulation-Info-Logic/src/main/java/org/apromore/processsimulation/service/SimulationInfoService.java
brucenguyen1: I see the need of making some names configurable via .properties file, e.g. "CUSTOM_TIMETABLE", "Default Timeslot". We're mixing both properties reader object and config object in one SimulationInfoConfig. One way of doing this without the above issue is create a separate properties reader object, read properties values into that object via Spring annotation, then pass that object to SimulationInfoConfig constructor, where it reads the properties and assigns non-empty property value to private variable or a private constant if the property is empty.

Apromore-Custom-Plugins/Process-Simulation-Info-Logic/src/main/java/org/apromore/processsimulation/service/SimulationInfoService.java
lahiru-maha: Wouldn't the code be safer this way though, instead of throwing a null pointer exception, that is incurred by the `@NonNull` annotation?

If a SimulationData object is created with a null calendar, we then need to catch the null pointer exception and either
- Show a message to the user or
- Default to using the 24/7 calendar

Either way I see the need for null handling

Apromore-Custom-Plugins/Process-Simulation-Info-Logic/src/main/java/org/apromore/processsimulation/service/SimulationInfoService.java
lahiru-maha: If no custom calendar is set, PD seems to use the calendar created via `CustomCalendarService.getGenericCalendar()` which in turn uses the `CalendarModel` class and not the `AbsoluteCalendarModel` so if we were to go down this route, we'd need to make some changes to the calendar model package, which I think is better achieved via its own task.
Let me know if my understanding here is incorrect.

Apromore-Custom-Plugins/Process-Simulation-Info-Logic/src/main/java/org/apromore/processsimulation/service/SimulationInfoService.java
lahiru-maha: I have made the change to this to simplify the loading of properties, and avoid having to use maps. Will address the immutability of the class in a later PR

Apromore-Custom-Plugins/Process-Simulation-Info-Logic/src/main/java/org/apromore/processsimulation/config/SimulationInfoConfig.java
lahiru-maha: I will address the immutability of the properties in a subsequent PR. For now, simplified the properties by removing the usage of maps

Apromore-Custom-Plugins/Process-Simulation-Info-Logic/src/main/java/org/apromore/processsimulation/service/SimulationInfoService.java
lahiru-maha: moved the hardcoded name to a constant.

Apromore-Custom-Plugins/Process-Simulation-Info-Logic/src/main/java/org/apromore/processsimulation/service/SimulationInfoService.java
brucenguyen1: Ok.

Apromore-Custom-Plugins/Process-Simulation-Info-Logic/src/main/java/org/apromore/processsimulation/config/SimulationInfoConfig.java
brucenguyen1: OK

Apromore-Custom-Plugins/Process-Simulation-Info-Logic/src/main/java/org/apromore/processsimulation/service/SimulationInfoService.java
brucenguyen1: Ok, it seems a little more work to do with CustomCalendar, we can do that later.

Apromore-Custom-Plugins/Process-Simulation-Info-Logic/src/main/java/org/apromore/processsimulation/service/SimulationInfoService.java
brucenguyen1: If there's a processing error in PD (simulation data provider) that makes it provide a null calendar (i.e. no calendar) while it does has a business calendar (i.e. 9-5 calendar), then the above error handling can't detect that error because it automatically "fixes" it by using a different calendar. In this contract (SimulationData), the CalendarModel is not optional (i.e. it can't be null) because the SimulationData comes with other time-related data such as average resource time, etc which is calculated based on the simulation data provider's calendar.

In error handling, it is not necessary to write error handling code for all types of errors/exceptions which could occur everywhere, the code will look more and more complicated and unrealistic. NullPointerException is a run-time error meaning something unexpected and wrong happens in the system. It can be handled via user-friendly message such as "Contact your administrator for help", error logs, and then fixed later. 

Usually, NPE (and run-time errors, e.g. divided by zero) should be covered as much as possible via unit tests/integration test, and then acceptance tests. But if some errors of this type still slip through, then system logs and incident handling process will be used.

brucenguyen1 (approved):

PR #1834, opened 2022-03-18
AP-6162 Ordering of folders and files in tree and tile/list view.

shapon2004:


raboczi (approved):
Tidy!  Possibly AlphaNumericComparator might even be worth sharing more widely via Apromore-Commons.

PR #1849, opened 2022-03-21
Addressing percentages not adding up to 100% due to rounding errors

lahiru-maha:
Addressing gateway probability percentages not adding up to 100% due to rounding errors
- if total % is less than 100 then add the difference to the edge with the lowest percentage
- if total % is greater than 100 then subtract the difference from the edge with the highest percentage

### DIFF
--- a/Apromore-Custom-Plugins/Process-Simulation-Info-Logic/src/main/java/org/apromore/processsimulation/service/SimulationInfoService.java
+++ b/Apromore-Custom-Plugins/Process-Simulation-Info-Logic/src/main/java/org/apromore/processsimulation/service/SimulationInfoService.java
@@ -310,8 +310,8 @@ public class SimulationInfoService {
 
                 // Calculate the total outbound edge frequencies for each gateway
                 double totalFrequency = gatewayEntry.getValue().stream()
-                    .map(EdgeFrequency::getFrequency)
-                    .reduce(0.0D, Double::sum);
+                    .mapToDouble(EdgeFrequency::getFrequency)
+                    .sum();
 
                 // Set the percentage for each edge's frequency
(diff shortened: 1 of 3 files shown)


Apromore-Custom-Plugins/Process-Simulation-Info-Logic/src/main/java/org/apromore/processsimulation/service/SimulationInfoService.java
```diff
@@ -306,11 +313,37 @@ private void deriveGatewayProbabilities(
...
 
+                // Set the percentage for each edge's frequency
+                gatewayEntry.getValue().forEach(edgeFrequency ->
+                    edgeFrequency.setPercentage(BigDecimal.valueOf(edgeFrequency.getFrequency() / totalFrequency)
+                        .setScale(4, RoundingMode.HALF_UP).doubleValue()));
+
+                // Determine if the percentages add up to a 100%
+                double totalProbabilities = gatewayEntry.getValue().stream()
+                    .map(EdgeFrequency::getPercentage)
+                    .reduce(0.0D, Double::sum);
```
raboczi: Perhaps [DoubleStream.sum](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/stream/DoubleStream.html#sum()) instead?

lahiru-maha: Yeah. Thanks for that. much cleaner 👍

raboczi (approved):
Honestly I'm a little skeptical of the idea of making the results less accurate in order to make them to appear more consistent, but that's an issue with the requirements rather than the implementation.  I have no other issue with merging this.

PR #1865, opened 2022-03-24
AP-6174: Improve edge point editing 

brucenguyen1:
Improve:
- Allows to edit edge points along its path
- Retain bezier curve during the editing (not to turn into straight line)

image

Caveat: the edge points can be edited during animation but token paths keep following the old path. Edge point editing should be disabled in animation mode but the cytoscape plugin hasn't provided API to update its state dynamically. This is for future work as we need to clone the code and customize it.

Details of changes: see commits below.


dreamind (approved):

dreamind:
@brucenguyen1 Could you merge the latest dev to this PR? I will then update the package-lock.json

brucenguyen1:
@dreamind : merged with latest development branch.

PR #1870, opened 2022-03-24
Ap-6223 Download multiple bpmn models corrupts the file.

shapon2004:


Janite (approved):
Works as expected. Was able to download a single model, multiple models and a combination of models and logs without the models being corrupted.

PR #1879, opened 2022-03-28
Feature/ap 5957 secure api with keycloak

jamesjin2021:
This PR provides method level access support for REST-API endpoints when Keycloak is enabled,  but also can be used for other modules. It is related to another ApromoreEE PR and Keycloak PR#153 https://github.com/apromore/keycloak-docker/pull/153.


raboczi (approved):
I haven't done functional tests, but I can't argue with the tidiness of code that's nothing but annotations.

PR #1900, opened 2022-03-31
AP-CHORE Permission fix

dreamind:
This is not strictly required for 8.3, it's just a convenience fix for developer to avoid switch db for changing permission set between dev and 8.3 branch.


Janite (approved):

Apromore-Clients/portal-model/src/main/java/org/apromore/portal/model/PermissionType.java
```diff
@@ -57,7 +57,8 @@
     MERGE_MODELS("7e1f8cc2-7532-406c-b106-3bf7095047dc", "Merge models"),
     SEARCH_MODELS("7c0f5dc2-7d39-456e-b1cb-139bb030ee98", "Search similar models"),
     PUBLISH_MODELS("607b5dfe-9508-11ec-b909-0242ac120002", "Publish models"),
-    UNREGISTERED("", "");
+    REST_API_ACCESS("0decac8d-b9a2-4424-8174-aace22ba4f39","Access rest apis"),
```
Janite: Can we rename this to REST_APIS so it matches the PermissionType in development?

Janite (changes requested):

Janite (approved):

PR #1901, opened 2022-03-31
bugfix/AP-6308 Prevent create user with same name as group

Janite:
This PR fixes a bug where an error would show if a user with the same name as a group was created and the group was selected. Added a check to prevent a user from being created if a group exists with the same name.


franklma (approved):
LGTM

PR #1904, opened 2022-04-01
AP-6225 Change ROLE_SUPER_ADMIN to ROLE_SUPERUSER in db

Janite:
This PR updates the name of ROLE_SUPER_ADMIN to ROLE_SUPERUSER in the db.

Please note: This will affect developers using the same database for both development and release/v8.3. Namely, the following will happen if using the same database for both versions:
- The ROLE_SUPERUSER is not authorized to use the REST APIs in release/v8.3. Any users affected by the name change will not be able to access v8.3 APIs.
- The ROLE_SUPERUSER will appear in the list of roles shown in the User Management window. "Super Admin" will not appear.

This PR has a pair in EE: https://github.com/apromore/ApromoreEE/pull/1277


raboczi (approved):
Does what it claims to do.

PR #1906, opened 2022-04-01
AP-6285 Fix model delete logic when it is created by PD and doesn’t h…

franklma:
…ave its draft branch created


shapon2004 (approved):
Looks fine!

PR #1908, opened 2022-04-01
AP-6248 Check existence of resources

dreamind:


shapon2004 (approved):
If you could remove the unused import. There might be conflict with baselistboxcontroller.java. Otherwise it works fine!

PR #1912, opened 2022-04-02
AP-6272 Add maxsize to the left panel

dreamind:


Janite (approved):
Looks good. The left panel is prevented from expanding past a point that causes the icons to overlap.

PR #1913, opened 2022-04-04
Refactoring of ZK libraries

raboczi:
This PR removes the redundant zkspring-core library.


franklma (approved):
Compile and run without issue. Tested main features, look OK to me.

PR #1916, opened 2022-04-04
Updating to springboot 2.5.12 in order to mitigate CVE-2022-22965 (v8.0)

lahiru-maha:


raboczi (approved):
Builds and runs both as part of EE and as standalone Core.

PR #1937, opened 2022-04-11
hotfix/v8.3/AP-6367 fix role migration result 

Janite:
- Add the analyst role to all users which only have the integrator role
- Remove the integrator role from all users with the superuser/super_admin role (roleid = 1)


franklma (approved):
works as expected

PR #1944, opened 2022-04-12
AP-6376 Version Sorting alphanumerically

shapon2004:


Janite (approved):
Works as expected. A model with versions [1.0, 2.0, 10.0] has versions sorted alphanumerically and version 10.0 selected as the latest version by default.

PR #1946, opened 2022-04-13
AP-6296: Apply new edge editing settings to trace/variant graphs, the same as the main graph

brucenguyen1:
This PR fixes a bug due to forgotten application of new edge editing settings for trace/trace variant graph visualization. Without the new settings, the visualization will show curve edges as seen below because it can't convert the edge property values properly.

Wrong:
image

Correct:
image


Janite (approved):
Looks good! Works as expected.

PR #1948, opened 2022-04-14
AP-6334 Fix updating of icon index

Janite:
This PR replaces .data('icon-index') with .attr('data-icon-index') so that the attribute 'data-icon-index' is updated in the DOM and can be used as a selector.


brucenguyen1 (approved):
Good fix: data(), attr() and using inline attribute = value in HTML elements, should be aware of the difference.

PR #1949, opened 2022-04-14
hotfix/v8.3/AP-6334 Fix update of icon index

Janite:
This PR replaces .data('icon-index') with .attr('data-icon-index') so that the attribute 'data-icon-index' is updated in the DOM and can be used as a selector.


brucenguyen1 (approved):
Good fix: there's a difference between data() and attr() method in jQuery, though it's confusing because data() can also be used to read attribute value.

PR #1952, opened 2022-04-19
feat: added env variables for the version of jars

mayankgituni:
Added environment input to take the version on the jar file. Using this we can do some version control in the dockerfiles and CI/CD.


lahiru-maha (approved):
Nice one

raboczi:
This doesn't makes sense to me.  The version number should be telling you what source code (and possibly what build configuration/profile) was used to produce the built artifact.  That makes it proper for it to be hardcoded into the source files.  If it's something that has to be hand-set by the builder, we introduce human error.

PR #1958, opened 2022-04-20
AP-6110: Set default cost to 0

brucenguyen1:
Default cost is set to 0, improve code for use of the default cost, add test coverage for this case.


franklma (approved):
works as expected

PR #1965, opened 2022-04-21
AP-6414 Apply calendar btn disabled after editing calendar

Janite:
Fixed a bug where a calendar was tagged as the applied calendar after editing it.


shapon2004 (approved):
Looks good!

PR #1967, opened 2022-04-21
Adding the skeletal structure for the event log ingestor for the new data layer.

lahiru-maha:
Adding the skeletal structure for the event log ingestor for the new data layer.
- New Log-Importer-Logic-EE module to contain the new implementation
- Conditional (property driven) bean that extends the LogImporter interface
- Minor formatting of some classes in the Apromore Storage module
- Adding tests, checkstyle, spotbugs and jacoco coverage

### DIFF
--- a/Apromore-Calendar/gradle.properties
+++ b/Apromore-Calendar/gradle.properties
@@ -1,2 +1,0 @@
-# plugins & build tools
-checkstyleVersion=9.2.1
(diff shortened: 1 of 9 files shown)


raboczi (approved):
Tempting to remove CsvConfig entirely in favor of just adding the file extensions as qualifiers to the various LogImporter-implementing Spring beans, but that'd be scope creep.

PR #1972, opened 2022-04-21
AP-6379 Simulation: Add arrival timetable dropdown in the General tab

shapon2004:


lahiru-maha (approved):
Code looks good. Tested locally and works as expected.

PR #1974, opened 2022-04-25
AP-6545-6455 Fix cost table state management

dreamind:


Janite (changes requested):
Works as expected according to the card. I've added a suggestion to change the currency_hint message since the current message implies that users can type in the box.

Apromore-Custom-Plugins/Process-Discoverer-Portal-Plugin/src/main/resources/static/processdiscoverer/zul/costTable.zul
```diff
@@ -60,6 +60,7 @@
       </label>
       <combobox
           disabled="${arg.viewOnly}"
+          readonly="true"
```
Janite, line 63: Since this box is now readonly, we should also change the currency_hint message to reflect that. The current message says "Choose a currency from the list or type in if not available", implying that a user can type in the box.

Janite (approved):

PR #1976, opened 2022-04-26
AP-5460 Several enhancements

franklma:
1. Only consider non-empty values when guessing the type by reading the 100 rows sample used in Log Importer
2. The “types” could be made optional and by default, when they are not given, all fields will be treated as “String”. With this “default” convention, maybe we would provide more flexibility to the API.
3. Corner cases when detecting the data type
4. Handle "other timestamps" in existing schema mapping
5. Handle Boolean data type
6. Add unit tests


Apromore-Custom-Plugins/Log-Importer-Logic/src/main/java/org/apromore/service/logimporter/model/LogMetaData.java
```diff
@@ -23,20 +23,27 @@
...
+import org.eclipse.collections.impl.list.Interval;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 
 @Data
 @NoArgsConstructor
 public class LogMetaData {
 
+    private static final Logger LOGGER = LoggerFactory.getLogger(LogMetaData.class);
```
mayankgituni: Could use Lombok notation @Sl4j for logging instead of defining LOGGER.

Apromore-Custom-Plugins/Log-Importer-Logic/src/main/java/org/apromore/service/logimporter/utilities/ImporterStringUtils.java
```diff
@@ -94,7 +97,12 @@ public static ColumnType getColumnType(final List<String> stringList) {
         temp.put(ColumnType.DOUBLE, 0);
         temp.put(ColumnType.TIMESTAMP, 0);
 
-        stringList.forEach(st -> {
+        List<String> filteredList = stringList.stream()
+            .filter(Objects::nonNull)
```
mayankgituni, line 101: Could combine both using StringUtils.isEmpty(). This will test for both the null and empty strings

mayankgituni (approved):
looks good to me!

PR #1978, opened 2022-04-27
feat: added the portal context and event utils

mayankgituni:
This PR moves the EventUtils from the ApromoreEE to ApromoreCore portal commons service. I also refactored the storeMetadataAsJSON to decouple from the portalContext.


Janite (approved):
Approving the Core PR but it should be merged with the EE one otherwise the ApromoreEE build will fail.

PR #1999, opened 2022-05-05
AP-6228: Set default servlet session timeout to 45m

raboczi:
This PR changes the default servlet session timeout to 45 minutes, from Spring Boot's default of 30.  This is longer than Keycloak's session length for JWT tokens (also 30 minutes), so it should be possible to reauthenticate from an expired JWT without losing the application state for a tab.


shapon2004 (approved):
I expect that it will solve the sudden reset issue.

PR #2001, opened 2022-05-05
AP-6486 Export attachment

dreamind:


Janite (commented):
Attachments added above the model are being cut off when we download it.

Janite (approved):
Looks good! The attachments are showing when we download svgs and pdfs.

PR #2036, opened 2022-05-16
AP-6610 Ability to change the default role of newly created users

shapon2004:


Janite (approved):
Looks good! Works as expected.

PR #2040, opened 2022-05-17
AP-6634 secure swagger ui page and use server default url for protoco…

jamesjin2021:
This PR aims to address the following 2 issues:

1. secure Swagger UI page and OpenAPI Json file so that only authenticated user can access to them.
2. use server default url for protocol scheme on Swagger UI page so that https is displayed for  https://springboot.apromore.org/swagger-ui/index.html#/


raboczi (approved):
A tidily prepackaged implementation.

The only issue I see is that redirection back to the Swagger page doesn't happen automatically while running without Keycloak.  The hassle to fix that doesn't seem warranted to me, since that's not a production configuration.

PR #2044, opened 2022-05-18
AP-5171: Override the Keycloak Spring Boot plugin

raboczi:
This PR overrides the generation of the login page URL by the Spring Boot plugin for Keycloak.  The overriding implementation is identical to the Keycloak 14 implementation, so there's no change in function yet, just the ability to change it in future.

Note that the code smells reported here are in code copied from Keycloak.  The reasoning is that it's better to minimize the changes than to fix them and introduce doubt that it does the same thing.


raboczi:
Superseded by https://github.com/apromore/ApromoreCore/pull/2176

PR #2049, opened 2022-05-18
AP-5171: Bookmarkable URL for Keycloak login page

raboczi:
Beware: this PR modifies its parent PR https://github.com/apromore/ApromoreCore/pull/2044 rather than the development branch!

This PR modifies the URL used for the Keycloak login page, adding `response_mode=fragment` and removing `state` from the query parameters of the OAuth URL to do this.


raboczi:
This didn't turn out to work.

PR #2061, opened 2022-05-20
AP-6283 Add notification after create new model

franklma:


dreamind (approved):

franklma:
Use another solution instead, hence close this one.

PR #2064, opened 2022-05-22
AP-6608 Hyperlinks support for PDF

dreamind:


Janite (changes requested):
There is a merge conflict in bpmneditor.js but otherwise, works as expected.

Janite (approved):

PR #2065, opened 2022-05-23
AP-6405 Import subprocesses as linked processes

Janite:
This PR introduces the following change:
When a user imports a model that contains a subprocess, the subprocess is imported as a seperate model linked to the original model.

The user flow is as follows:
1. Upload a BPMN model which contains at least 1 subprocess.
2. Multiple processes should be added to the workspace. The subprocess models will have _subprocess1, _subprocess2 etc. appended to the end of the model name.
3. Open imported model in BPMNEditor. Clicking on the [+] on a subProcess should open the linked subprocess model in a new tab.


dreamind (approved):

Apromore-Core-Components/Apromore-Portal/src/main/java/org/apromore/portal/dialogController/ImportOneProcessController.java
```diff
@@ -154,6 +173,76 @@ public void importProcess(final String domain, final String owner)
...
+    Map<ProcessSummaryType, Map<BPMNNode, BPMNNode>> subProcessModelClonedNodesMap = new HashMap<>();
+    int count = 0;
+    for (SubProcess subProcess : bpmnDiagram.getSubProcesses()) {
+      BPMNDiagram subProcessDiagram = BPMNDiagramFactory.newBPMNDiagram("");
+      Map<BPMNNode, BPMNNode> nodeMap = subProcessDiagram.cloneSubProcessContents(subProcess);
+      String subProcessName = name + "_subprocess" + ++count;
+      ProcessSummaryType subProcessSummary = importSubProcess(subProcessDiagram, domain, owner, subProcessName, folderId);
+      subProcessMap.put(subProcess, subProcessSummary);
+      subProcessModelClonedNodesMap.put(subProcessSummary, nodeMap);
+    }
```
brucenguyen1: As discussed, we need to make this code easier to understand by using better naming, or it could be better with explicit class/data type rather than all using maps.

Apromore-ProcessMining-Collection/prom-bpmn/src/main/java/org/apromore/processmining/models/graphbased/directed/bpmn/BPMNDiagramImpl.java
```diff
@@ -280,6 +281,52 @@ protected Map<DirectedGraphElement, DirectedGraphElement> cloneFrom(
...
+	public Map<BPMNNode, BPMNNode> cloneSubProcessContents(SubProcess subProcess) {
+		BPMNDiagram bpmnDiagram = (BPMNDiagram) subProcess.getGraph();
+		Map<BPMNNode, BPMNNode> nodeMap = new HashMap<>();
+
+		//Nodes
+		for (ContainableDirectedGraphElement subProcessChild : subProcess.getChildren()) {
+			if (subProcessChild instanceof BPMNNode) {
+				BPMNNode node = (BPMNNode) subProcessChild;
+				nodeMap.put(node, node.addCopyToDiagram(this));
+			}
```
brucenguyen1: As discussed, this code is actually: first we need "node" to return a copy BPMNNode from itself called copyNode, then "this" BPMN diagram will add this copyNode to itself. The code can be outlined below. The method BPMNNode.addToDiagram is a bit shortcut and circle because it actually creates a copy from its _own constructor_ and the call the diagram to add this copy.

BPMN copyNode = node.copy().
this.addNode(copyNode).

BPMNNode.copy() and BPMNDiagram.addNode() need to be implemented.
Another issue is in the BPMNDiagram.addNode method, atm, it has to do if (node instanceof XXX) where XXX is all types of nodes. We should be looking for an improvement for this if...else block.

Apromore-Core-Components/Apromore-Portal/src/main/java/org/apromore/portal/dialogController/ImportOneProcessController.java
```diff
@@ -154,6 +173,76 @@ public void importProcess(final String domain, final String owner)
...
+        + "xmlns:dc='http://www.omg.org/spec/DD/20100524/DC' "
+        + "targetNamespace='http://bpmn.io/schema/bpmn' " + "id='Definitions_1'>"
+        + "<bpmn:process id='Process_1' isExecutable='false'/>"
+        + "<bpmndi:BPMNDiagram id='BPMNDiagram_1'>"
+        + "<bpmndi:BPMNPlane id='BPMNPlane_1' bpmnElement='Process_1'/>"
+        + "</bpmndi:BPMNDiagram>"
+        + "</bpmn:definitions>";
+
+    String bpmnText = bpmnDiagram.getNodes().isEmpty() ? emptyModelXML : BpmnLayoutPlugin.addLayout(bpmnDiagram, "");
+    ImportProcessResultType importResult = mainC.getManagerService().importProcess(owner,
```
brucenguyen1: As discussed, this PR needs a batch import operation because it imports multiple diagrams (each for one sub-process) and they are all related. Any error happens will leave the database in an unknown state with some garbage sub-processes already imported. Imagine, it would be a nightmare to clean up the database if it happens to a large digaram with many sub-processes.

Apromore-Core-Components/Apromore-Portal/src/main/java/org/apromore/portal/dialogController/ImportOneProcessController.java
Janite: A tree structure has been introduced to make this part easier to understand.

Apromore-Core-Components/Apromore-Portal/src/main/java/org/apromore/portal/dialogController/ImportOneProcessController.java
Janite: As discussed, the batch import will be handled as part of a further improvement, seperate from this PR. Marking this comment as resolved.

brucenguyen1 (approved):

PR #2069, opened 2022-05-23
AP-6691: Don't secure PortalPluginResource URLs

raboczi:
This PR adjusts the Spring Security configuration for both the Keycloak and non-Keycloak modes such that static content served via the PortalPluginResourceServlet (fonts, icons, etc) isn't subject to access control.  This is similar to the way all other static content is currently treated.


dreamind (approved):
Works as expected

PR #2070, opened 2022-05-23
AP-6283 Add model created notification

franklma:


Apromore-Core-Components/Apromore-Portal/src/main/java/org/apromore/portal/dialogController/BPMNEditorController.java
```diff
@@ -155,8 +155,11 @@ public BPMNEditorController() {
     process = session.getProcess();
     vst = session.getVersion();
 
+    Map<String, Object> param = new HashMap<>();
+
     if (isNewProcess) {
       currentUserAccessType = AccessType.OWNER;
+      param.put("isNewProcess", true);
```
Janite, line 162: This parameter is missing when we open an existing model in BPMNEditor or we open a model using a publish link which prevents the BPMNEditor from opening.
We can move it out of the if case so it exists when opening an existing model and also add it in pushViewModeParameters() so it exists when opening a model using a publish link.

When we open a new model, it works as expected.

Janite (changes requested):

Janite (approved):

PR #2079, opened 2022-05-25
fix/avoid error caused by the missing runtime object

chiichang:
- avoid the error caused by the missing runtime object when calling the method from UnitTest


shapon2004 (approved):
Looks good!

PR #2088, opened 2022-05-25
AP-6719 Fix color palette for label

dreamind:


Janite (approved):
Black is now available as a color option for edges.

PR #2094, opened 2022-05-26
INC-25: Version bump of ZK to version 10 preview

raboczi:
This PR requires its counterpart https://github.com/apromore/ApromoreEE/pull/1484 from ApromoreEE in order to build correctly.  It bumps the ZK framework version to 10.0.0.FL.20220413.  This is a preview version available from the ZK Maven repository but not from Maven Central yet.  The Iceblue Compact theme remains at version 9.6, since it doesn't seem to (yet?) exist in the new version.
* The main feature of ZK 10 is alternative stateless UI components. See [ZK10 Preview: ZK 10 is ready for building Cloud-Native Applications](https://www.zkoss.org/wiki/Small_Talks/2022/April/ZK10_Preview:_ZK_10_is_ready_for_building_Cloud-Native_Applications).  Existing ZUL-based component continue to work.
* Also includes the security patch addressing INC-25 we've been waiting for in ZK 9.6.2

Upon cursory examination, everything seems to be back-compatible with our existing ZK 9.6 code, but it seems like a good idea to ask for people with deeper familiarity with the UI to confirm this.  (Update: an incompatibility has indeed been found, and https://github.com/apromore/ApromoreEE/pull/1484 created to address it.)

Since this is just a version bump in the Gradle config, it may be better to just look at the [Files changed](2094/files) and do the edits by hand rather than check out this branch and have to worry about whether it's still synced with ApromoreEE.


shapon2004 (approved):
Application is working fine. All looks good!

raboczi:
This turned out to have too many side effects.

PR #2095, opened 2022-05-26
AP-6741 Select current linked process on edit subprocess link

Janite:
This PR contains the following fixes related to linking subProcesses:

1. When the "Link subprocess model" window is opened, the current linked model should now as the selected model. The "Link subprocess model" window can be opened by opening the Properties panel in bpmn editor with a subprocess selected, and clicking "Edit" in the Linked subprocess section.
Screen Shot 2022-05-26 at 1 15 15 pm

2. Added a check to prevent the view link from opening if the user does not have access to the linked model.


shapon2004 (commented):
"throws UserNotFoundException" - I think throwing exception on @init method could expose error stacktrace to the client. It is better to catch it locally and destroy the related desktop/component if required.

shapon2004 (approved):
Works as expected!

PR #2104, opened 2022-05-30
AP-6702: Add Content-Security-Policy HTTP header

raboczi:
This PR configures Spring Boot to include a Content-Security-Policy HTTP header specified as the contentSecurityProperty property of the application context.

It also removes some other Spring Security configuration of the HTTP headers, generally ones which Spring Boot already sets by default.  A change is that the constraint on HTTP frames has been tightened from "same origin" to "deny" (X-Frame-Origin header, and CSP frame-ancestors 'none') since I don't think we employ frames.

This PR will need to be backported to releases 8.3 and 8.4, but I'd like it to go through QA first.

### DIFF
--- a/Apromore-Boot/src/main/resources/application.properties
+++ b/Apromore-Boot/src/main/resources/application.properties
@@ -6,6 +6,15 @@ server.servlet.session.cookie.http-only=true
 server.servlet.session.cookie.cookie.path=/
 server.servlet.session.timeout=45m
 
+# See https://content-security-policy.com for format details
+contentSecurityPolicy=default-src 'self';\
+ font-src 'self' data: fonts.googleapis.com fonts.gstatic.com;\
+ form-action 'self';\
+ frame-ancestors 'none';\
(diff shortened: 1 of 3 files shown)


dreamind (approved):

dreamind:
Security Policy directive: "img-src 'self' will block all images linked from the BPMN model attachment. I perhaps need to disable external linking. Will confirm with @marcellolarosa

raboczi:
> Security Policy directive: "img-src 'self' will block all images linked from the BPMN model attachment. I perhaps need to disable external linking. Will confirm with @marcellolarosa

If we need to make the policy more liberal, at least it's in the configuration now rather than in the code.

raboczi:
It turns out that the log importer uses frames internally, so there's a follow-up PR https://github.com/apromore/ApromoreCore/pull/2109 to relax the frame-ancestors CSP directive and the X-Frame-Origin header back to 'same'/SAMEORIGIN.

PR #2108, opened 2022-05-30
Bugfix/ap 6744/cant import export parquet log

Chill-X:
Changes:
- Ensure unique temp parquet files when exporting from XES


mayankgituni (approved):
Looks good!

PR #2123, opened 2022-06-01
Introducing the Process Log Definition Service

lahiru-maha:
- Introducing new Database tables, entities and surrounding services to better save and retrieve the ProcessLog  definition (i.e metadata) via the Process Log Manager module
- Update the new log importer to persist the new metadata (this logic will later be moved to the ProcessLogService itself)


franklma (approved):
LGTM

Apromore-Database/src/main/java/org/apromore/dao/model/LogicalLogAttribute.java
```diff
@@ -0,0 +1,79 @@
...
+
+@Entity
+@Table(name = "logical_log_attribute",
+    uniqueConstraints = {
+        @UniqueConstraint(columnNames = {"id"})
+    }
+)
+@Configurable("logical_log_attribute")
+@Getter
+@Setter
```
raboczi: You've removed setters from the other entities.  Was this one retained intentionally?

Apromore-Database/src/main/java/org/apromore/dao/model/BaseEntity.java
```diff
@@ -0,0 +1,49 @@
...
+import javax.persistence.GeneratedValue;
+import javax.persistence.Id;
+import javax.persistence.MappedSuperclass;
+import lombok.Getter;
+import lombok.Setter;
+import org.hibernate.annotations.CreationTimestamp;
+import org.hibernate.annotations.UpdateTimestamp;
+
+@Getter
+@Setter
```
raboczi: Maybe only the "updated_timestamp" should be settable?  Alternatively, maybe there shouldn't even be a setter for that, only be an "update" method that sets updated_timestamp to the current time.

lahiru-maha: Missed removing the setters from base entity. will do that now

Apromore-Database/src/main/java/org/apromore/dao/model/LogicalLogAttribute.java
lahiru-maha: Need one setter to set the bi directional relationship with PhysicalLogAttribute. Will remove the high level setter but keep the one specific to the PhysicalLogAttribute

Apromore-Database/src/main/java/org/apromore/dao/model/DataType.java
```diff
@@ -0,0 +1,51 @@
...
+ * These values were chosen to match https://www.xes-standard.org/_media/xes/xesstandarddefinition-2.0.pdf §2.2,
+ * which also includes some we’re not using: “boolean”, “id”, “list”.
+ */
+@JsonNaming(PropertyNamingStrategies.SnakeCaseStrategy.class)
+public enum DataType {
+    STRING,
+    DATE,
+    DOUBLE,
+    INT;
+
```
brucenguyen1: It looks like this DataType is meant to cover the most common primitive data found for data attributes, right? Following this line, I wonder why long is not included here. For example, timestamp is all long values and indispensable in our data, not sure if the intention here is to cover it with String or Double, but these two are not suitable (String is unchecked type, and Double is lost precision).

raboczi (approved):
All my quibbles have been addressed.

Apromore-Database/src/main/java/org/apromore/dao/model/DataType.java
raboczi: These values are taken directly from https://apromore.atlassian.net/wiki/spaces/AP/pages/2143748188/Apromore-Log-Meta+header+format with the exception that Lahiru figured "float" from XES would confuse people in Java, since it's different from Java's single-precision `float` type.  I guess you're arguing similarly against XES's "int".

Apromore-Database/src/main/java/org/apromore/dao/model/DataTypeDao.java
```diff
@@ -0,0 +1,48 @@
...
+@Configurable("data_type")
+@Getter
+@Builder
+@NoArgsConstructor
+@AllArgsConstructor
+public class DataTypeDao extends BaseEntity {
+    @Column(name = "type")
+    @Enumerated(EnumType.STRING)
+    private DataType type;
+}
```
brucenguyen1: If we want to use Builder for creating these objects manually, should we include Builder.Default to make it null safe (avoid NPE at runtime)? For example, the createdTimestamp could be null after calling builder.build() without caling builder.createdTimestamp(), and so the created object is left in an unknown state with null createdTimestamp.

Same comment for other objects with Builder, like LogicalLogAttribute.

Apromore-Database/src/main/java/org/apromore/dao/model/DataTypeDao.java
```diff
@@ -0,0 +1,48 @@
...
+        @UniqueConstraint(columnNames = {"id"})
+    }
+)
+@Configurable("data_type")
+@Getter
+@Builder
+@NoArgsConstructor
+@AllArgsConstructor
+public class DataTypeDao extends BaseEntity {
+    @Column(name = "type")
```
brucenguyen1: It's a bit confusing when only this particular entity is named <EntityName>Dao while others are not with the suffix Dao while they are the same data access objects. DataType could be renamed DataTypeEnum because it is only one, and keeps this one DataType for entities like the others.

Apromore-Database/src/main/java/org/apromore/dao/model/PhysicalLog.java
```diff
@@ -0,0 +1,55 @@
...
+@Builder
+@NoArgsConstructor
+@AllArgsConstructor
+public class PhysicalLog extends BaseEntity {
+
+    @OneToOne(mappedBy = "physicalLog")
+    Log log;
+
+    @Column(name = "hive_table_name")
+    private String hiveTableName;
```
brucenguyen1: Shouldn't use implementation-specific  for data model. Hive is considered as a catalog in Presto/Trino data model. MySQL can be another catalog. Maybe catalog_table_name is more generic.

Apromore-Database/src/main/java/org/apromore/dao/model/DataType.java
lahiru-maha: I've been pondering this as well. should we expand the DataTypes to include others such as long, timestamp? I briefly discussed this with Marlon as well, and we're going to do the same in the converters as well (which currently only supports `int`, `string`, `date` and `float`.

Given that when we convert these basic types to the Hive schema types we will be translating
`int` -> `bitint` 
`float` -> `double`
`string` -> `varchar(255)`
`date` -> `timestamp`

I believe there would be an advantage in providing a richer set of datatypes. We then of course need to update the REST API as well to make use of these

Apromore-Database/src/main/java/org/apromore/dao/model/DataTypeDao.java
lahiru-maha: Yeah. I went for this since the Enum is what will be used across the other modules when setting types, and not so much the dao. The dao would only be used by the persistence layer when mapping between the enum to the dao. Thought the code will read better as DataType instead of DataTypeEnum.

Apromore-Database/src/main/java/org/apromore/dao/model/PhysicalLog.java
lahiru-maha: Yeah that's a good point. Will change this

Apromore-Database/src/main/java/org/apromore/dao/model/DataTypeDao.java
lahiru-maha: The createdTimestamp and updatedTimestamp are never manually set. These are taken care of by the respective hibernate annotations.

But your comment is valid for some other attributes (eg. index, attributeTypeDao, dataTypeDao in LogicalLogAttribute). Will check how to add these safety checks

Apromore-Database/src/main/java/org/apromore/dao/model/DataType.java
lahiru-maha: I had already intentionally changed `float` to `double` in the DataTypes because of this

Apromore-Database/src/main/java/org/apromore/dao/model/DataTypeDao.java
brucenguyen1: Another suggestion is: TypeEnum is shorter and relevant and leave DataType for entity. If you read the code here, you're a little bit confused with DataTypeDao next to non-Dao already, just read it again right above and below :)

lahiru-maha: The only reason you see those two together at the moment is because of the the Apromore EE / Core constraint. As mentioned the other day, I will be moving the enums into an seperate package within the process log definition module, and the dao entities into a seperate model package. but this work depends on moving Apromore database from the open source repo to the closed source repo.

Couldn't really use TypeEnum cause we have AttributeType and DataType

lahiru-maha: Also another reason I avoided suffixing an enum with `Enum` at the end was based on this https://stackoverflow.com/questions/3069743/coding-conventions-naming-enums

Apromore-Database/src/main/java/org/apromore/dao/model/DataType.java
brucenguyen1: From our discussion, I understood this DataType is high level and not one-one mapped to specific types (like Java, SQL). But the naming is quite specific: Double, Int, String (this looks like Java, e.g. why calling it "Double"). A better naming is: String = Text (or Literal), Double = Number_Continuous, Int = Number_Discreet.

brucenguyen1: DataType.DATE here should be "DATETIME". There's a missing type for time only. Java has OffsetTime, and MySQL has TIME data type (https://dev.mysql.com/doc/refman/8.0/en/time.html). In CustomCalendar module, it uses time data only for workday (e.g. from 9am to 5pm).

lahiru-maha: As discussed, going with `STRING, TIMESTAMP, REAL, INTEGER;` as the high level data types. and also to be in line with https://apromore.atlassian.net/browse/AP-5460

PR #2144, opened 2022-06-05
AP-6790 Search giving an error but still displaying the results

shapon2004:


Janite (approved):
Search works as expected

PR #2146, opened 2022-06-06
AP-6703 Fix js vulnerabilities in Apromore Frontend

dreamind:


Janite (approved):
Bpmn editor, pd, log animation and conformance checking appear to be working as expected.

PR #2148, opened 2022-06-06
Hotfix/ap 6744/cant import export parquet log

Chill-X:


mayankgituni (approved):
Approving but I couldn't test this due to the schema changes in the Apromore core. We will merge all the PRs first and then test the fixes collectively.

PR #2157, opened 2022-06-09
AP-6404 Chain linked subprocesses during export and add checks for circular references

Janite:
- Check for circular process links when attempting to link a subprocess or export a model.
- Include linked subprocesses of linked subprocesses during export instead of only the first level of linked subprocesses.
- Generate random ids for linked subprocess elements included in the exported bpmn file.

To verify:
1. Attempt to link a process to itself or in a circular fashion (e.g. A -> B -> C -> A). An error message should appear.
2. Link multiple subprocesses in a chain (e.g. link model A -> model B -> model C via subprocess), export with linked subprocesses included and re-import the exported model. Multiple models should be imported and linked in the same chain as the exported model.


shapon2004 (approved):
Looks fine and works as expected mentioned into Jane's comment. Multiple models are exported and also point to correct model.

PR #2160, opened 2022-06-09
AP-6447 Discover hourly cost information from event log and use in si…

shapon2004:
…m parameters


shapon2004:
deleting it because, i want to upload a fresh copy.

PR #2161, opened 2022-06-10
AP-6807 Fix simulation paremeter numerical input

dreamind:


Janite (approved):
Invalid values can be typed in a number field and show an error without changing to NaN.

PR #2163, opened 2022-06-10
AP-6447 Discover hourly cost information from event log and use in si…

shapon2004:
…m parameters

### DIFF
--- a/Apromore-Custom-Plugins/Process-Simulation-Info-Logic/src/main/java/org/apromore/processsimulation/service/SimulationInfoService.java
+++ b/Apromore-Custom-Plugins/Process-Simulation-Info-Logic/src/main/java/org/apromore/processsimulation/service/SimulationInfoService.java
@@ -324,9 +324,6 @@ public class SimulationInfoService {
 
     private Map<String, Double> retrieveCostData(int logId) {
         Map<String, Double> roleToCostRateMap = new HashMap<>();
-        if (objectMapper == null) {
-            return roleToCostRateMap;
-        }
         try {
             objectMapper.configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false);
             List<Usermetadata> userMetadata =
@@ -342,8 +339,8 @@ public class SimulationInfoService {
                 roleToCostRateMap = costingDataList[0].getCostRates();
             }
         } catch (JsonProcessingException ex) {
-            log.warn("Error in parsing cost table usermetadata for log {}", logId);
-        } catch (Exception ex) {
+            log.warn("Error in parsing cost usermetadata for log {}", logId);
+        } catch (RuntimeException ex) {
             log.warn("Error in retrieving cost data for log {}", logId);
         }
         return roleToCostRateMap;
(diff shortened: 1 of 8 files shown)


Apromore-Custom-Plugins/Process-Simulation-Info-Logic/src/main/java/org/apromore/processsimulation/service/SimulationInfoService.java
```diff
@@ -293,21 +306,46 @@ private Map<String, String> deriveResourceInfo(
...
+                return costRateMap;
+            }
+            CostingData[] costingDataList =
+                objectMapper.readValue(userMetadata.get(0).getContent(), CostingData[].class);
+            if (costingDataList != null && costingDataList.length > 0
+                && costingDataList[0].getCostRates() != null) {
+                costRateMap = costingDataList[0].getCostRates();
+            }
+        } catch (JsonProcessingException ex) {
+            log.error("Error in json data parsing", ex);
```
lahiru-maha: Might be better to be a bit more specific with the error message. i.e. "Error in parsing cost json data". Also since we are not erroring out of the overall operation, I reckon a log.warn() would be more suitable here

Apromore-Custom-Plugins/Process-Simulation-Info-Logic/src/main/java/org/apromore/processsimulation/service/SimulationInfoService.java
```diff
@@ -293,21 +306,46 @@ private Map<String, String> deriveResourceInfo(
...
+        } catch (Exception ex) {
+            log.error("Error in retrieving data", ex);
```
lahiru-maha: Same as above.. "Error in retrieving cost data" might read better. And similarly, since we are not erroring out of the overall operation, I reckon a log.warn() would be more suitable here

Apromore-Custom-Plugins/Process-Simulation-Info-Logic/src/main/java/org/apromore/processsimulation/service/SimulationInfoService.java
```diff
@@ -84,11 +91,16 @@ public class SimulationInfoService {
 
     private final SimulationInfoConfig config;
     private final CalendarService calendarService;
+    private UserMetadataService userMetadataService;
+    private ObjectMapper objectMapper;
 
     @Autowired
-    public SimulationInfoService(SimulationInfoConfig config, CalendarService calendarService) {
+    public SimulationInfoService(SimulationInfoConfig config, CalendarService calendarService,
+                                 UserMetadataService userMetadataService, ObjectMapper objectMapper) {
```
lahiru-maha: I'm wondering if passing the object mapper in as a bean to be auto wired is a good idea. This would mean that if someone else is making use of the same object mapper, they could potentially change its properties, which in turn may affect the deserialisation behaviour here.

Would it be safer to create an instance of ObjectMapper within this constructor instead.
i.e. in line 103
`this.objectMapper = new ObjectMapper();`

Apromore-Custom-Plugins/Process-Simulation-Info-Logic/src/main/java/org/apromore/processsimulation/service/SimulationInfoService.java
```diff
@@ -293,21 +306,46 @@ private Map<String, String> deriveResourceInfo(
...
                         .build();
                 }).collect(Collectors.toList()));
         }
 
         return resouceNameToId;
     }
 
+
+    private Map<String, Double> retrieveCostData(int logId) {
+        Map<String, Double> costRateMap = new HashMap<>();
```
lahiru-maha: I would rename the variable to be `roleToCostRateMap` for better readability

lahiru-maha (approved):
LGTM. Thanks for addressing the review comments

PR #2171, opened 2022-06-15
BPMN Sub-process related problems resolved

shapon2004:
Following cards fo BPMN editor has been consided into this commit: 

AP-6741 - Editor: when editing subprocess link, current selection is lost (Checked- works fine)
Ap-6875 - When user links a new process, even after saving the name of the linked process it still shows as untitled
Ap-6884 - Clicking on link subprocess window throws an error
Ap-6879 - The folders that are not shared with the user are displayed after choosing the 'use existing process model'


Janite (approved):
Looks good!
- [x] AP-6741 - Current selection remains on edit
- [x] AP-6875 - Refreshing a model linked from "Create new" shows the linked model's updated name
- [x] AP-6884 - No error appears after double clicking the Edit button or [+] icon
- [x] AP-6879 - The link subprocess model window only shows folders in the user's workspace

PR #2173, opened 2022-06-15
BPMN Sub-process related problems resolved

shapon2004:


shapon2004:
Following cards fo BPMN editor has been consided into this commit:

AP-6741 - Editor: when editing subprocess link, current selection is lost (Checked- works fine)
Ap-6875 - When user links a new process, even after saving the name of the linked process it still shows as untitled
Ap-6884 - Clicking on link subprocess window throws an error
Ap-6879 - The folders that are not shared with the user are displayed after choosing the 'use existing process model'

Janite (approved):
Works as expected.

PR #2176, opened 2022-06-16
Configure Keycloak login for proxy

raboczi:
This PR allows the OAuth2 redirect URL used by the Keycloak login process to be configured via the `portal.login-url` Spring property.  This allows the externally-visible HTTPS URL to be used so the the AWS ALB doesn't have to insert corrective redirections.


raboczi:
Note that the SonarCloud issues apply to code copied verbatim from the Keycloak 14 codebase.  I think it's best to modify this code as little as possible so that it's easier to distinguish the changes from the orginal.

raboczi:
This needs to be redone as part of the ApromoreEE repository instead.

PR #2179, opened 2022-06-16
Ap-6886 Role is unchecked for User B after unchecking role for User A…

shapon2004:
… when both users are selected


Janite (approved):
Looks good. Works as expected.

PR #2180, opened 2022-06-17
AP-6864 Turn autodisable buttons on to prevent dblclick

dreamind:


Janite (approved):
Double clicking items in the portal toolbar no longer causes the the non-unique id message to appear.

PR #2182, opened 2022-06-17
Setting the corresponding date format from the LogMetaData on each applicable ProcessLogAttribute

lahiru-maha:


franklma (approved):
LGTM

PR #2185, opened 2022-06-20
AP-6922 Fix bug when linking to a process in a loop

Janite:
This PR fixes a bug where a StackOverflowError would appear when linking a subprocess to a process that was already linked with a circular reference (e.g. models A -> B -> C -> A are linked and user attempts to link D -> A).


shapon2004 (approved):
Looks Good.

PR #2187, opened 2022-06-21
AP-6937 Fix 2MB payload limit for zkau HTTP requests

dreamind:


Apromore-Boot/src/main/resources/application.properties
```diff
@@ -147,3 +147,16 @@ server.error.whitelabel.enabled=false
...
+server.tomcat.max-http-post-size=100000000
+spring.servlet.multipart.max-file-size=100MB
+spring.servlet.multipart.max-request-size=100MB
+spring.servlet.multipart.maxFileSize=100MB
+spring.servlet.multipart.maxRequestSize=100MB
+spring.http.multipart.max-file-size=100Mb
+spring.http.multipart.max-request-size=100Mb
+multipart.maxFileSize=100Mb
+multipart.maxRequestSize=100Mb
+server.tomcat.max-http-form-post-size=100000000
```
raboczi: I wonder if it's possible to do something like the following so that changing the value will be less onerous and error-prone:
```
maxRequestMB = 100
server.tomcat.max-swallow-size=${maxRequestMB}000000
server.tomcat.max-http-post-size=${maxRequestMB}000000
spring.servlet.multipart.max-file-size=${maxRequestMB}MB
```
(etc)

dreamind:
@raboczi we've decided to keep only the one that works on the springboot server.

raboczi (approved):

PR #2189, opened 2022-06-21
AP-6881 add log permission check for log downloading

jamesjin2021:
This PR adds a method canUserReadLog() for checking if an user has read permission on a specified log in event log service.  Such check is used for log downloading API to prevent user from download logs not owned or no read permission.  There is another ApromoreEE PR#1569 depends on this PR. https://github.com/apromore/ApromoreEE/pull/1569

### DIFF
--- a/Apromore-Core-Components/Apromore-Manager/src/main/java/org/apromore/service/impl/EventLogServiceImpl.java
+++ b/Apromore-Core-Components/Apromore-Manager/src/main/java/org/apromore/service/impl/EventLogServiceImpl.java
@@ -475,6 +475,15 @@ public class EventLogServiceImpl implements EventLogService {
 	return false;
     }
 
+	@Override
+	public boolean canUserReadLog(String username, Integer logId) throws UserNotFoundException {
+		User user = userSrv.findUserByLogin(username);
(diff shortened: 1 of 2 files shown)


Apromore-Core-Components/Apromore-Manager/src/main/java/org/apromore/service/impl/EventLogServiceImpl.java
```diff
@@ -475,6 +475,13 @@ public boolean canUserWriteLog(String username, Integer logId) throws UserNotFou
 	return false;
     }
 
+	@Override
+	public boolean canUserReadLog(String username, Integer logId) throws UserNotFoundException {
+		User user = userSrv.findUserByLogin(username);
+		return groupLogRepo.findByLogAndUser(logId, user.getRowGuid()).stream()
+			.anyMatch(gl -> gl.getAccessRights().isReadOnly() || gl.getAccessRights().isOwnerShip());
```
raboczi, line 482: My goodness, the accessor method names on the `AccessRights` class are dangerously confusing (isReadOnly rather than hasReadPermission).  But this looks correct despite that.

raboczi (approved):
I feel like the AccessRights class desperately needs renaming, but that's not part of this PR.

PR #2192, opened 2022-06-21
AP-6943 Fix NPE when removing access to a folder that contains an unshared artifact

Janite:
Fix NPE that appeared when removing user access to a folder that contains models or logs not shared with the removed user.


franklma (approved):
H2 table "storage" which uses the reserved word "key" as identifiers would fail the integration test. This affects the test coverage but should not block this PR. We can revisit this issue later https://apromore.atlassian.net/browse/AP-6961

PR #2200, opened 2022-06-22
AP-6920 Hide case variant column in case details

dreamind:


shapon2004 (approved):
Looks like column will not be used anymore. In that case, we could remove it.

PR #2204, opened 2022-06-23
AP-6764: Bump ZK version to 9.6.0.2

raboczi:
This PR bumps the version of ZK from 9.6.0 to 9.6.0.2.  This incorporates the security fix for issue ZK-5150 (see https://tracker.zkoss.org/browse/ZK-5150).


Janite (approved):
Looks good! Did a sanity check and nothing seems to break.

PR #2208, opened 2022-06-26
AP-6907 Global search implementation for portal

dreamind:


raboczi (approved):
I wish this code had tests, but the change itself seems tidy enough.

dreamind:
> I wish this code had tests, but the change itself seems tidy enough.

I will create a separate PR for test. Mocking up the comprehensive fixture seems to be a bit complicated

PR #2212, opened 2022-06-28
AP-6952 Add option to download models with linked subprocesses in portal

Janite:
This PR adds an option for users to include linked subprocesses in models downloaded from the main portal (either by clicking download in the toolbar with models selected or by selecting download from the right-click menu with at least one model selected).


shapon2004 (approved):
Works as expected!

PR #2213, opened 2022-06-28
AP-4695 Fix svg, bpmn, pdf filename export

dreamind:


Janite (approved):
The model name is used when downloading a bpmn, svg or pdf from bpmn editor as expected. Added the missing license header.

PR #2217, opened 2022-06-28
Ap-6446 Sim parameters: limit total number of instances to be simulated

shapon2004:


Janite (approved):
The number of process instances in the simulation params is capped at the value specified by `process-simulation-info-export.default-max-process-instances`.

PR #2219, opened 2022-06-29
AP-6936 Copy subprocess links when copying a model (v8.4)

Janite:
This PR introduces a change to copy subprocess links when copying a model. The subprocesses in a copied model should link to the same process models as the original when the model is created.


franklma (approved):
LGTM

PR #2223, opened 2022-07-01
AP-6991 Fix bpmn editor and diff title

dreamind:


Janite (approved):
Looks good. Ellipsis are added if the model name is too long.

PR #2224, opened 2022-07-01
Feature/ap 6962 find logs by name and folder and user group core

jamesjin2021:
This PR adds a method findReadableOrOwnedLogsByNameInFolderForUser() in class 'LogRepository.java' in module 'Apromore-Database' for finding readable logs in a folder for an user.  This method is used by log downloading API to check if the user can read the downloaded log. There is a related ApromoreEE PR https://github.com/apromore/ApromoreEE/pull/1603 depends on this PR.


raboczi (approved):
Functional testing will confirm whether it's correct, but it's certainly safe.

PR #2233, opened 2022-07-07
AP-7125 Fix PD export PDF

dreamind:


Janite (approved):
Added license header.

Export pdf in PD is showing the whole model, as expected.

PR #2239, opened 2022-07-15
AP-7158 Automatically open an unsaved draft

Janite:
This PR removes the prompt asking users whether they would like to open an unsaved draft or disregard the changes. Instead, the draft is opened automatically.


franklma (approved):
LGTM

PR #2241, opened 2022-07-15
AP-7169-7125 Fix PD label and large size export

dreamind:


raboczi (approved):
Looked fine before, still looks fine now.

PR #2246, opened 2022-07-19
AP-4695 Fix svg, bpmn, pdf filename export

dreamind:


Janite (approved):
Looks good. Added the missing license header to bpmneditor.js

PR #2254, opened 2022-07-22
AP-7197 Change PDF export from svg-based to image-based

dreamind:


dreamind:
The ServletRegistrationBean approach is not consistent with other Spring-based REST API implementation. This would be addressed in different ticket.

raboczi (approved):
This isn't in development, so I'm not going to overly fussy.  Good enough!

PR #2256, opened 2022-07-23
AP-7254 Fix distro values conversion for sim params

dreamind:


shapon2004 (approved):
`if (value === 'NaN') {
        value = ''
    }`
Empty value might give error in simulator engine. Would you please check that.

PR #2263, opened 2022-07-26
AP-7175 Open home folder when showing search result

Janite:
Open the home folder by default when showing or clearing search results in process model search.


shapon2004 (approved):
Looks good!

PR #2264, opened 2022-07-26
AP-7202 Add predictive monitoring permissions

Janite:
This PR adds role permissions for predictive monitoring. These permissions can be viewed in via User Management > Role > View role.

The following roles should have predictive monitoring permissions:

- Superuser, Analyst, Data Scientist: both Manage predictors and View predictions
- Operations: only View predictions


franklma (approved):
looks good to me

PR #2266, opened 2022-07-26
AP-7184 Correction on extended Simulation parameter

shapon2004:


shapon2004:
AP-7266 The "Probability must not exceed 100%" error is displayed after clicking the "Simulate model" button
AP-7184 The “Sum of probabilities must be equal to 100%“ message is displayed all time
AP-7252 User is able to save generated simulation log if Sum of probabilities aren't equal to 100% for some categorical case attributes
AP-7260 The "Probability must not exceed 100%," message is displayed after entering more than 100%

dreamind (approved):

PR #2270, opened 2022-07-28
AP-7293 Add new KC user default role based on prop value

dreamind:


shapon2004 (approved):
It looks good to me.But I think Simon's review is also required here.

raboczi (approved):
Looks fine to me.  There's one minor tweak SonarCloud picked up which seems worthwhile.

Apromore-Clients/portal-model/src/main/java/org/apromore/portal/model/RoleName.java
```diff
@@ -0,0 +1,36 @@
...
+ *
+ * You should have received a copy of the GNU General Lesser Public
+ * License along with this program.  If not, see
+ * <http://www.gnu.org/licenses/lgpl-3.0.html>.
+ * #L%
+ */
+
+package org.apromore.portal.model;
+
+public final class RoleName {
```
raboczi, line 25: SonarCloud suggests a private constructor here.

PR #2275, opened 2022-07-28
AP-7023 NPE: Saving a BPMN file including simulation params (Release 8.4)

shapon2004:


shapon2004:
Resolved into another commits.

PR #2287, opened 2022-08-04
AP-7263 Add option to unpublish linked subprocesses when publishing model 

Janite:
This PR adds a prompt which allows users to unpublish linked subprocesses when publishing a model with the "publish linked subprocesses" option unchecked.


shapon2004 (approved):
looks good to me!

PR #2288, opened 2022-08-05
AP-7352 Update default behaviour of publish linked-subprocesses checkbox

Janite:
Updated the isLinkedProcessPublished() method to include indirect linked subprocess models in the check.

The "include linked subprocesses" checkbox in the Publish model window should now be checked by default if the model contains a published indirect linked subprocess.
(i.e. for models linked A -> B -> C, if model C is published and model B is not published, model A should have the checkbox selected by default).


shapon2004 (approved):
Looks good!

PR #2292, opened 2022-08-09
Feature/ap 4242 Loganimation is now possible for models with pools, data objects, text annotations and implicit gateways

reissnda:
-The class process selection encapsulates pre- and post processing steps for process models to make them valid for the log-animation plugin.
-Log-animation can only handle one pool. Process selection will select the first pool that contains a start and end event.
-Log-animation can not parse text annotations, data objects, association flows. Process selection will only retain elements that correspond to the control flow, i.e. events, tasks, sequence flows and gateways.
-Tasks or events with multiple incoming (xor-join) or outgoing (and-split) sequence flows represent implicit gateways. Process selection adds model elements in a pre-processing step to explicitly model these gateways for the replay and removes these added elements in a post-processing step for the animation.
-All functionality contains tests in the ProcessSelectionTest class.
-Additionally, this pull request also contains a bug fix for AP-7157.


brucenguyen1 (approved):
Built and ran with no issues.

PR #2301, opened 2022-08-17
AP-7436 change pd (old DL) shortcut to open node cost filter with cost overlay active (v8.4.1)

Janite:
This PR has a pair in ApromoreEE: https://github.com/apromore/ApromoreEE/pull/1904

This PR updates the shortcut in PD such that, if Shift + Right-click is performed on a node with the cost overlay active the  node cost filter will be opened  instead of node duration.

Other changes: Fixed test failures for Apromore-Frontend in release/v8.4.1.


Apromore-Custom-Plugins/Process-Discoverer-Portal-Plugin/src/main/java/org/apromore/plugin/portal/processdiscoverer/impl/apmlog/LogFilterControllerWithAPMLog.java
```diff
@@ -151,14 +152,21 @@ private Map<String, Object> getEditorOptionPayload(FilterType filterType,
...
+            case NODE_COST:
                 data = (String) param.get("data");
                 if (filterType == FilterType.EVENT_ATTRIBUTE_DURATION &&
                         !isValidAttributeDurationPayload(mainAttribute, data)) {
                     Messagebox.show(
                             parent.getLabel("failedFilterNodeDurationSingleValue_message"),
-                            parent.getLabel("common_error_text"), Messagebox.OK, Messagebox.ERROR);
+                            parent.getLabel(ERROR_TITLE_KEY), Messagebox.OK, Messagebox.ERROR);
                     return null;
                 }
```
brucenguyen1: This condition is not used (always false) because the filterType is NODE_COST in this switch case.

Apromore-Custom-Plugins/Process-Discoverer-Portal-Plugin/src/main/java/org/apromore/plugin/portal/processdiscoverer/impl/apmlog/LogFilterControllerWithAPMLog.java
```diff
@@ -151,14 +152,21 @@ private Map<String, Object> getEditorOptionPayload(FilterType filterType,
...
+                if (filterType == FilterType.NODE_COST && parent.getProcessAnalyst().getRoleValues().isEmpty()) {
+                    Messagebox.show(
```
brucenguyen1: Should check parent.getProcessAnalyst().getRoleValues().isEmpty() is enough, the filterType is being NODE_COST in this switch case.

Apromore-Custom-Plugins/Process-Discoverer-Portal-Plugin/src/main/java/org/apromore/plugin/portal/processdiscoverer/impl/apmlog/LogFilterControllerWithAPMLog.java
brucenguyen1: ah I see. This is for three switch cases, not one.

Janite: This code is reached if filterType is either CASE_SECTION_ATTRIBUTE_COMBINATION,  EVENT_ATTRIBUTE_DURATION or NODE_COST so there is still a need to check.

Apromore-Custom-Plugins/Process-Discoverer-Portal-Plugin/src/main/java/org/apromore/plugin/portal/processdiscoverer/impl/apmlog/LogFilterControllerWithAPMLog.java
brucenguyen1: I see, this is for 3 switch cases, not one. So it is if then else again within switch case.

Janite: This code is reached if filterType is either CASE_SECTION_ATTRIBUTE_COMBINATION,  EVENT_ATTRIBUTE_DURATION or NODE_COST so there is still a need to check for the filterType.

Apromore-Custom-Plugins/Process-Discoverer-Portal-Plugin/src/main/java/org/apromore/plugin/portal/processdiscoverer/impl/apmlog/LogFilterControllerWithAPMLog.java
brucenguyen1: Can we move the condition check to its own switch case and no need to check filterType again?

brucenguyen1: Can we move the error message handling outside and before the switch case? It is a separate part from the switch case handling. It's not easy to read switch...case and then if...then...else again within switch case (a bit circular).

Apromore-Custom-Plugins/Process-Discoverer-Portal-Plugin/src/main/java/org/apromore/plugin/portal/processdiscoverer/impl/apmlog/LogFilterControllerWithAPMLog.java
brucenguyen1: The same comment, can we move the error message first and outside the switch case to make the switch...case block cleaner to read?

Apromore-Custom-Plugins/Process-Discoverer-Portal-Plugin/src/main/resources/static/processdiscoverer/zul/processDiscoverer.zul
```diff
@@ -134,6 +134,11 @@
                         'ap-mc-icon-play',
                         'ap-mc-icon-pause');
 
+    Ap.pd.primaryOverlay = 'FREQUENCY';
+    Ap.pd.setPrimaryOverlay = function(measureType) {
+      Ap.pd.primaryOverlay = measureType;
+    }
```
brucenguyen1: Can we move this function definition to processdiscoverer/graph.js instead of making it lonely here while the rest of all PD functions are there? Putting some functions in the zul file makes it hard to manage PD code, where it uses Ap.pd.primaryOverlay === 'COST', but magically setting value for Ap.pd.primaryOverlay is located in a ZUL file.

brucenguyen1: Can we make primaryOverlay part of the _private variable set of PD? This is to indicate that it is private.

image

brucenguyen1: In the _private variable set above, for primaryOverlay, we can set it to 'FREQUENCY' in PD constructor (the same as other instance variables). It is more consistent than setting it alone after the constructor in the ZUL file.

brucenguyen1 (approved):

PR #2303, opened 2022-08-18
AP-7226 Disable cost overlay and cost centre if no cost info is defined in pd (old DL) (v8.4.1)

Janite:
This PR has a pair in ApromoreEE: https://github.com/apromore/ApromoreEE/pull/1918

This PR makes the following changes to pd (old DL):
- If the log opened in pd has no roles defined, cost overlay and cost centre are disabled.
- If the log opened in pd has roles but no costs assigned (all the costs are zero), selecting cost overlay instead opens a popup which prompts users to set costs in the cost centre.

### DIFF
--- a/Apromore-Custom-Plugins/Process-Discoverer-Portal-Plugin/src/main/java/org/apromore/plugin/portal/processdiscoverer/components/ToolbarController.java
+++ b/Apromore-Custom-Plugins/Process-Discoverer-Portal-Plugin/src/main/java/org/apromore/plugin/portal/processdiscoverer/components/ToolbarController.java
@@ -29,7 +29,6 @@ import org.apromore.plugin.portal.processdiscoverer.data.ContextData;
 import org.apromore.plugin.portal.processdiscoverer.data.UserOptionsData;
 import org.apromore.service.AuthorizationService;
 import org.apromore.util.AccessType;
-import org.eclipse.collections.api.list.ImmutableList;
 import org.zkoss.zk.ui.Component;
 import org.zkoss.zk.ui.event.Event;
 import org.zkoss.zk.ui.event.EventListener;
--- a/Apromore-Custom-Plugins/Process-Discoverer-Portal-Plugin/src/main/java/org/apromore/plugin/portal/processdiscoverer/components/ToolbarController.java
+++ b/Apromore-Custom-Plugins/Process-Discoverer-Portal-Plugin/src/main/java/org/apromore/plugin/portal/processdiscoverer/components/ToolbarController.java
@@ -197,16 +197,7 @@ public class ToolbarController extends AbstractController {
     }
 
     public void setDisabledCost(boolean disabled) {
-        boolean hasNoRoles = false;
-        try {
-            ImmutableList<Object> roleValues = parent.getProcessAnalyst().getRoleValues();
-            if (roleValues.isEmpty()) {
-                hasNoRoles = true;
-            }
-        } catch (Exception e) {
-            hasNoRoles = true;
-        }
-
+        boolean hasNoRoles = parent.getProcessAnalyst().getRoleValues().isEmpty();
         cost.setDisabled(disabled || hasNoRoles);
     }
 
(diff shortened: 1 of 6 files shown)


Apromore-Custom-Plugins/Process-Discoverer-Portal-Plugin/src/main/java/org/apromore/plugin/portal/processdiscoverer/components/ToolbarController.java
```diff
@@ -193,6 +196,20 @@ public void setDisabledAnimation(boolean disabled) {
...
+    public void setDisabledCost(boolean disabled) {
+        boolean hasNoRoles = false;
+        try {
+            ImmutableList<Object> roleValues = parent.getProcessAnalyst().getRoleValues();
+            if (roleValues.isEmpty()) {
+                hasNoRoles = true;
+            }
+        } catch (Exception e) {
+            hasNoRoles = true;
+        }
```
brucenguyen1: There could be other exception thrown out in case of errors but has been silenced. The PDAnalyst.getRoleValues should handle null like what has been done in https://github.com/apromore/ApromoreCore/pull/2301. If so, it doesn't need to catch any exception here.

Apromore-Custom-Plugins/Process-Discoverer-Portal-Plugin/src/main/java/org/apromore/plugin/portal/processdiscoverer/components/ViewSettingsController.java
```diff
@@ -189,6 +192,18 @@ public void initializeControls(Object data) {
...
         costAggSelector = (Combobox) compViewSettings.getFellow("costAggSelector");
 
+        try {
+            ImmutableList<Object> roleValues = parent.getProcessAnalyst().getRoleValues();
+            if (roleValues.isEmpty()) {
+                costDisabled = true;
+            }
+        } catch (Exception e) {
+            costDisabled = true;
+        }
```
brucenguyen1: Same comment as above.

Apromore-Custom-Plugins/Process-Discoverer-Portal-Plugin/src/main/java/org/apromore/plugin/portal/processdiscoverer/components/ViewSettingsController.java
```diff
@@ -144,6 +146,7 @@ public class ViewSettingsController extends VisualController {
     private static final String ON_FORCE_SELECT = "onForceSelect";
 
     private boolean disabled = false;
+    private boolean costDisabled = false;
```
brucenguyen1: Maybe better naming: costCentreNotAvailable?

Apromore-Custom-Plugins/Process-Discoverer-Portal-Plugin/src/main/java/org/apromore/plugin/portal/processdiscoverer/components/ViewSettingsController.java
```diff
@@ -248,10 +263,14 @@ public void initializeEventListeners(Object data) {
...
 
         defaultCost.addEventListener("onClick", event -> {
-                if (disabled) return;
-                String key = "mean";
-                selectComboboxByKey(costAggSelector, key);
-                selectCostViz();
+            if (disabled || costDisabled) return;
+            if (userOptions.getCostTable().isDefault()) {
+                showNoCostSetDialog();
+                return;
```
brucenguyen1: It is more accurate to check all costs are zero, not default cost because the default cost can be non-zero (it might be changed later). The meaning here is all zeros means the cost table is not set as described in the Jira card.
It could be safer if we implement a CostTable.isNotSet method where it checks for all zeros, and use this method instead of isDefault. Because if somebody changes the default value to non-zero then this implementation won't work correctly any more.

Apromore-Custom-Plugins/Process-Discoverer-Portal-Plugin/src/main/java/org/apromore/plugin/portal/processdiscoverer/components/ViewSettingsController.java
Janite: Changed the variable name to isCostCentreAvailable and updated the checks + assignments.

brucenguyen1 (approved):

PR #1983, opened 2022-04-28
Implementation of the file upload service

lahiru-maha:
Implementation of the file upload service
Adding tests to the Storage and Storage-EE services
Adding checkstyle and spotbugs to the Storage and Storage-EE services

PR #2320, opened 2022-08-30
AP-7444: Log REST users and HTTP request IDs (v8.5)

raboczi:
This PR is a back-port of https://github.com/apromore/ApromoreEE/pull/1983 to the release/v8.5 branch.  It provides request ID tracking in the log.  It also removes the PID column from the log, which is interfering with Gradle due to some manner of interaction with the `$PID` variable when application.properties is copied.


jamesjin2021 (approved):
looks good, approve it.

PR #2323, opened 2022-09-05
AP-7686 Hide ppm and etl role permissions from user management if not enabled

Janite:
This PR hides the ppm/etl role permissions from the view/edit role permission window if ppm/etl is not enabled.


franklma (approved):
LGTM

PR #2324, opened 2022-09-05
AP-7650 Fix external label sizing

dreamind:


Janite (commented):
- Labels on events and gateways can be resized and text adjusts as expected.
- Labels on arcs are still not resizable.

Janite (approved):
Looks good. Arc, event and gateway labels can be resized and text adjusts as expected.

PR #2328, opened 2022-09-14
AP-7168-7585 Fix save into submitByEnter handler

dreamind:
Fix save into problem when user press Enter after putting in file name


shapon2004 (approved):
looks good!

PR #2333, opened 2022-10-03
AP-8016 Remove permissions from default roles

Janite:
This PR removes permissions from default roles according to AP-8016.


franklma (approved):
LGTM

PR #2334, opened 2022-10-06
AP-7998 [PD] Error when clicking twice on "save BPMN model" button

shapon2004:


Janite (approved):
Looks good. Error no longer shows when double clicking "save BPMN model" in pd

PR #2335, opened 2022-10-07
AP-7945 Update Japanese translations (v8.5)

Janite:
This PR updates the Japanese translation property files. Note: some properties are still in english because the translation is missing.

Other changes:
- Fix incorrect labelling of includeLinkedSubProcess_title and includeLinkedSubProcess_message property
- Move "Select a folder" label (from save into folder window) into a property file


shapon2004 (approved):
Looks good!

PR #2343, opened 2022-11-09
AP-8291 Update japanese labels for v8.5

Janite:
- Update the predictive monitoring labels for the sub menu controller
- Add Apromore.I18N.ShapeSize in translation_ja to avoid an exception when referencing Apromore.I18N.ShapeSize.shapeSize with Japanese as the selected language.


shapon2004 (approved):
Looks good!

PR #2356, opened 2023-03-07
AP-9298 Delete assets option does not work when deleting the user

franklma:
Delete orphan artifacts by invoking EventLogService, ProcessService and WorkspaceService when deleting user.


dreamind (approved):
Looks good.

PR #2358, opened 2023-04-13
fixed issue DatabaseException: Illegal mix of collations for operatio…

molihan:
…n 'UNION'

When the encode is missing you can't execute the sql successfully when init.


franklma:
> …n 'UNION'
> 
> When the encode is missing you can't execute the sql successfully when init.

Hi, @molihan. Thanks for your pull request. Could you explain how to reproduce this DatabaseException: Illegal mix of collations for operation 'UNION'? As mentioned in README [mysql-setup](https://github.com/apromore/ApromoreCore#mysql-setup) section, the database Apromore is created with the character set and collation specified. So all tables are supposed to inherit the same setting unless overwritten.

